PR #87, opened 2013-04-29
Twig.php: test/Twig/Tests/Fixtures/expressions/literals.test

m-abs:
Bools can be both upper and lower case.
Output of bools differ from PHP and JavaScript:
In PHP, true prints as "1", in JavaScript it prints as "true".
IN PHP, false prints as "", in JavaScript it prints as "false".


evgenius:
Hi @m-abs , thanks for creating this PR.
Sorry you had to wait so long.
I am going to merge it but then modify it a bit as Twig.php doesn't consider `TrUe` as `true`.
Only all lower/upper case is supported (see https://github.com/twigphp/Twig/blob/master/lib/Twig/ExpressionParser.php)

PR #101, opened 2013-06-15
Implement the Twig.php test iterable and add specialcase test for it

m-abs:
This adds the test 'iterable' to Twig.js.

In test-ext/twig.php/test/Twig/Tests/Fixtures/tests/iterable.test with test data is as follows:
```php
return array(
    'foo' => array(),
    'traversable' => new ArrayIterator(array()),
    'obj' => new stdClass(),
    'val' => 'test',
);
```

I've decided to use this test data:
```json
{
 'foo' : [],
 'traversable' : [],
 'obj' : {},
 'val' : 'test'
}
```
As JS don't have a widely implemented alternative to ArrayIterator. Mozilla mentions "Iterator" but it's not implemented in Node.js https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators


evgenius:
The test "iterable" was added in https://github.com/justjohn/twig.js/commit/6603d32.
@m-abs I'll try to merge your changes to `test-ext/test.original.js` later though, thank you.

PR #164, opened 2014-06-11
Implement max and min functions

jmealo:
**I'll preface this PR with this:**
The behavior of PHP's ``min`` and ``max`` functionals when passed associative arrays is ... *interesting*.

Regrettably, I've included the ``max`` and ``min`` functions from phpjs to ensure full compatibility. (The ``max`` and ``min`` functions are 90% similar and very long for what they do.)

There's much better ways to implement this using native ``Array.sort()`` and ``Math.max()`` and ``Math.min()``... however, the result doesn't behave as ``min`` and ``max``, so it won't pass the unit tests.

I didn't hack apart ``min`` and ``max`` to save a few bytes as gzip will take care of this. I would like to revisit any phpjs code we include to see if we can optimize some of the common code re-used between snippets.

### DIFF
--- a/demos/node_express/public/vendor/twig.js
+++ b/demos/node_express/public/vendor/twig.js
@@ -1802,7 +1802,228 @@ var Twig = (function(Twig) {
         }
 
         return (isHalf ? value : Math.round(value)) / m;
-    }
+    };
+
+    Twig.lib.max = function max() {
+        //  discuss at: http://phpjs.org/functions/max/
+        // original by: Onno Marsman
+        //  revised by: Onno Marsman
+        // improved by: Jack
+        //        note: Long code cause we're aiming for maximum PHP compatibility
+        //   example 1: max(1, 3, 5, 6, 7);
+        //   returns 1: 7
+        //   example 2: max([2, 4, 5]);
+        //   returns 2: 5
+        //   example 3: max(0, 'hello');
+        //   returns 3: 0
+        //   example 4: max('hello', 0);
+        //   returns 4: 'hello'
+        //   example 5: max(-1, 'hello');
+        //   returns 5: 'hello'
+        //   example 6: max([2, 4, 8], [2, 5, 7]);
+        //   returns 6: [2, 5, 7]
+
+        var ar, retVal, i = 0,
+            n = 0,
+            argv = arguments,
+            argc = argv.length,
+            _obj2Array = function (obj) {
+                if (Object.prototype.toString.call(obj) === '[object Array]') {
+                    return obj;
+                } else {
+                    var ar = [];
+                    for (var i in obj) {
+                        if (obj.hasOwnProperty(i)) {
+                            ar.push(obj[i]);
+                        }
+                    }
+                    return ar;
+                }
+            }; //function _obj2Array
+        _compare = function (current, next) {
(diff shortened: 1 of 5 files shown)


evgenius:
@jmealo Would you mind porting the tests from Twig.php?
https://github.com/twigphp/Twig/blob/master/test/Twig/Tests/Fixtures/functions/min.test
https://github.com/twigphp/Twig/blob/master/test/Twig/Tests/Fixtures/functions/max.test

jmealo:
@evgenius Wow, I'm surprised that I didn't do that!  I'll get those in at some point this weekend. These should pass no problem since they use phpjs for compatibility reasons already.

I'm not sure how I figured out that a more ideal implementation didn't match PHP functionality without those tests ;-) hm...

jmealo:
Based on the PR, I must have already had these coded and forgot to commit them.

evgenius:
@jmealo No problem :)
You saw my comment about `_compare` declared globally, didn't you?

jmealo:
@evgenius, yes, my mistake to think that phpjs was sane code...

evgenius:
@jmealo They already merged my PR so all is good :)

evgenius:
@jmealo So could you update your branch to fix the issues below or do you want me to do it? I won't change the authorship of the commit :)
1. _compare declared globally
2. Add tests from Twig.php
3. Exclude built files from the PR?

jmealo:
@evgenius I see that you have a ``_compare`` fix in php.js upstream.

If the ``_compare`` and ``obj2array`` are identical between the functions it seems like a shame to include them twice...

evgenius:
Wow, I missed it. Yeah, I think it's okay to declare it somewhere in twig.functions.js.
Say, after `function is`. Since we're going to split the original code, I think it's also okay to use `is('Array')` instead of the original `Object.prototype.toString.call(obj) === '[object Array]'`.

jmealo:
@evgenius  I'm pulling out ``_compare`` and ``obj2array`` and putting them in ``Twig.lib``. I don't like using the phpjs stuff, but when I did max/min without using them it was impossible to duplicate PHP functionality. I don't think that there are tests for ``max`` and ``min`` in Twig as the PHP version just calls PHP's ``max`` and ``min``. I can see if PHP has unit tests for them (I'm sure it does).

evgenius:
Do you think we need it in Twig.lib? It seems to be for the functions used widely in the whole library.

jmealo:
@evgenius I wondered the same thing... I guess we should remove it from there, and put it back in there if anything else uses them. I think that phpjs uses those functions in multiple places.

In reviewing the PHP Unit tests for this function, we already have adequate test coverage. However, the phpjs versions throw errors where PHP would have only issued warnings and returned somewhat erratic return values (sometimes null, sometimes false, but always "falsey") .

I do not think we should throw errors in places where the errors are considered warnings as this will break templates that just work in PHP.

jmealo:
I thought this would be quicker than it was. Converting ``max`` and ``min``  to return ``null`` and ``false`` instead of throwing errors causes weird return values in Twig.js.

The ``max`` and ``min`` behaving just as they do in PHP may not make them behave the same in Twig. I will have to do additional testing on the PHP side.

PR #166, opened 2014-06-12
Implement random function

jmealo:
This pull request implements a ``random`` function that matches the behavior of the ``twig_random`` function that appears here:
https://github.com/fabpot/Twig/blob/7c63a7df8a9d38ff9eb9fb982ae8f6a70258c10c/lib/Twig/Extension/Core.php#L386

### DIFF
--- a/src/twig.functions.js
+++ b/src/twig.functions.js
@@ -199,6 +199,8 @@ var Twig = (function (Twig) {
                 var keys = Object.keys(value);
                 return value[keys[getRandomNumber(keys.length-1)]];
             }
+
+            return getRandomNumber(2147483647);
         }
     };
 
--- a/src/twig.functions.js
+++ b/src/twig.functions.js
@@ -177,6 +177,28 @@ var Twig = (function (Twig) {
             return new Twig.Template({
                 data: template
             });
+        },
+        random: function(value) {
+            function getRandomNumber(n) {
+                return Math.floor((Math.random() * n) + 1);
+            }
+
+            if(Twig.lib.is("Number", value)) {
+                return getRandomNumber(value);
+            }
+
+            if(Twig.lib.is("String", value)) {
+                return value.charAt(getRandomNumber(value.length-1));
+            }
+
+            if(Twig.lib.is("Array", value)) {
+                return value[getRandomNumber(value.length-1)];
+            }
+
+            if(Twig.lib.is("Object", value)) {
+                var keys = Object.keys(value);
+                return value[keys[getRandomNumber(keys.length-1)]];
+            }
         }
     };
 
--- a/test/test.functions.js
+++ b/test/test.functions.js
@@ -261,5 +261,29 @@ describe("Twig.js Functions ->", function() {
                 .should.equal('test');
             });
         });
+
+        describe("random ->", function() {
+            it("should return a random item from a traversable or array", function() {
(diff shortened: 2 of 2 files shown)


jmealo:
@evgenius Good catch. It really depends whether @justjohn wants to implement Twig as it is documented or as it behaves in PHP. It appears that not passing parameters is undefined behavior. It's not mentioned in the [documentation](http://twig.sensiolabs.org/doc/functions/random.html) or in the unit tests.

In my opinion, Twig's behavior shouldn't be defined by PHP's built-in functions, but by Twig itself.

Should Twig implementations in other languages (and possibly future versions of PHP) be burdened to implement Twig and every PHP function that it calls?

Perhaps @justjohn can weigh in on where this project stands on the matter?

The code that I have contributed so far (minus this pull request apparently ;-)) leaned towards 100% compatibility.

evgenius:
Hi @jmealo ,

This is what @justjohn thinks about how close Twig.js should be to Twig.php:
> My goal is to mimic twig PHP as closely as possible.

The full comment is here: https://github.com/justjohn/twig.js/issues/88#issuecomment-40725877

justjohn:
@evgenius is correct, I'd prefer to keep compatibility with twig php where possible. That way it's possible to reuse existing templates written for twig php in twig js without having to rewrite them too much.

evgenius:
I remember there was my comment about `getRandomNumber(1)` returning `1` only.
Probably because of the rebase it's lost. Weird. Probably it happens when the PR is based on the fork's branch.

I took a look at the PHP source code and looks like `mt_rand` uses Mersenne Twister to generate a pseudo-random number.
Well, it would be ideal to use it but it seems to be overcomplicating of the code so let's leave this idea :)

So to make the function `getRandomNumber` work correctly its code should look like
```js
// returns a number in a range [0, n]
function getRandomNumber(n) {
    return Math.floor(Math.random() * (n + 1));
}
```

Though I would simplify it to not include `n`:
```js
// returns a number in a range [0, n)
function getRandomNumber(n) {
    return Math.floor(Math.random() * n);
}
```
Then its call would look like:
`arr[getRandomNumber(arr.length)]`
and
`getRandomNumber(val + 1)`
and
`getRandomNumber(0x80000000)` (when no parameters are passed)

evgenius:
Argh, then `random(-1)` wouldl never return `0`.

evgenius:
Ok, then let's take a look at PHP source code:
`PHP_RAND` at https://github.com/php/php-src/blob/PHP-5.4/ext/standard/php_rand.h
```c
#define RAND_RANGE(__n, __min, __max, __tmax) \
    (__n) = (__min) + (long) ((double) ( (double) (__max) - (__min) + 1.0) * ((__n) / ((__tmax) + 1.0)))
```
They just take a random number in [0, tmax] range and the interpolate it to a number in the [min, max] range.
So thus they don't care about positive and negative values, they're covered automatically.

In JS it will be the following:
```js
function getRandomNumber(n) {
    var LIMIT_INT31 = 0x80000000;
    var random = Math.floor(Math.random() * LIMIT_INT31);
    var limits = [0, n];
    var min = Math.min.apply(null, limits),
        max = Math.max.apply(null, limits);
    return min + Math.floor((max - min + 1) * random / LIMIT_INT31);
}
```

Issue #174, opened 2014-08-22
Errors are reported via console.log, why not use console.error ?


paulgreg:
Twig.log.error = function() {
       console.log.apply(console, arguments);
    }

When declaring [Twig.log.error](https://github.com/justjohn/twig.js/blob/master/src/twig.core.js#L150) function, why is it using console.log ?
Shouldn’t it be using console.error (if avaible) ?

It’s problematic in my case because I’ve got error reported by node in "out" log file and not into the "err" one.

evgenius:
Fixed via #175

PR #175, opened 2014-08-29
Issue #174 - use console.error instead of console.log

paulgreg:
As discussed in issue #174, we had to log errors in err.log file in our node.js app.
Here’s a pull request which fixes that.


evgenius:
Thanks @paulgreg !

PR #176, opened 2014-09-17
fix scoping issue in for loops

createvibe:
Copying the context object each time is creating scoping issues in the for loops.  By maintaining the reference to context, variables that are created and / or modified inside the loop still exist outside the loop as well.


evgenius:
Hi @createvibe ,

Thank you for your PR.
Unfortunately I can't accept it as there is an ongoing optimization of a for loop.
If you're curious, I created a PR already: #211 . It solves the issues with a loop scope.

Here's a couple notes about your code:
- respect the indentation style of the surrounding code
- no need to change the style of the variable name that was already used, i.e. inner_context -> innerContext.

In other words, try to keep others' code untouched unless you really need to rewrite it.
Looking forward getting other PRs from you.
Thank you!

PR #186, opened 2014-10-16
Allow int 0 as a key in objects

csimi:
This fails right now due to an incorrect truthy check.
I'm not sure if also allowing an empty string would be a good idea.


plepe:
Merged to master in 4528370f07eacdd117381b23dc24e8ae42680351 (using the modified branch from @evgenius)

PR #190, opened 2014-11-19
Add truncate filter

nike-17:
- port truncate filter from twig extention repo https://github.com/twigphp/Twig-extensions/blob/master/lib/Twig/Extensions/Extension/Text.php#L42
-  add unit test
-  build files


evgenius:
@nike-17 Thanks for your contribution.
@jmealo Thanks for checking it.

PR #195, opened 2014-12-17
Support to twig namespaces / bundle-name

Zizaco:
In order to get a better compatibility between twig being used in front-end and in the back-end (with PHP) we've made a fork in order to add "namespace" compatibility.

## Namespace

    ...
    var template = Twig.twig({
        data: breadcrumbTpl,
        namespaces: { 'my-namespace': 'path/to/views/' }
    }).render();
    ...

Done this, when referencing another template, instead of using the relative path, it is possible to use the _namespaces_ that were previously defined. Ex:

    {# random-template.twig #}
    {% extends "kameleon-skin::default.twig" %}
    ...

The `"my-namespace::"` will now point to `"path/to/views/"`.


jmealo:
@rafaellyra I don't think the coverage is tracked in twig.js.

I did a quick look for any tests in the PHP version covering the code that you implemented here and nothing jumped out at me. It's been a while since I reviewed this PR and may have missed it.

Either way, looks like it's time for a rebase! :-)

micahgodbolt:
Any chance this any closer to getting merged in? Evaluating twigjs and I really need this feature.

rafaellyra:
@micahgodbolt we are using our fork since the PR is open and it's working very well :+1:

rafaellyra:
Rebase done, a new PR was created here -> https://github.com/justjohn/twig.js/pull/251

PR #197, opened 2015-01-03
Fix travis builds

eiriksm:
Hey.

Just noticed travis build started failing for this repo. And noticed it was because some of the dependencies started declaring dependencies with carets instead of tilde. So, just upgrading npm before the script starts fixes that.

Unfortunately newer npm does not work on node 0.6, so in this PR I also removed node 0.6 from the testing. I am guessing not many people use that anymore.

Thanks for making this and making it available, looking forward to playing more with it!

Have a nice day!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: node_js
 node_js:
-  - 0.6
   - 0.8
   - 0.10
   - 0.11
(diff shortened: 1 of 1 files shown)


evgenius:
@eiriksm I was thinking about removing 0.6 from the list too but I wasn't sure.
Let's do it. Thanks for making it happen!

eiriksm:
Yay!
And the badge is green again. Good times!

Issue #210, opened 2015-02-16
Sync "set" tag behavior with Twig.php


evgenius:
Twig.php behavior:
- if a variable is declared in the for-loop scope, it stays undefined globally
- if a variable is declared globally and modified in the for-loop scope, changes should affect the global variable

evgenius:
I think I already fixed it.

Issue #191, opened 2014-11-25
Iteration problem in nested loops


meefox:
Looks like there is an issue with scoping variable in loops
```
{% set k = 0 %}
{% for i in 0..5%}
	{% for j in 0..5 %}
		{{ k }}
		{% set k = k + 1 %}
	{% endfor %}
{% endfor %}
```
I expect that it will iterate by 1 each time.
Like 0,1,2,3,4,5,6 and so on, but it is 0,0,0,0,0,0,1,1,1,1,1,1,2,2,2,2,2,2,3 and so on.

evgenius:
Tested in 0.8.2: works fine now.

Issue #144, opened 2014-03-02
Twig.js requires optimization of a "for" tag


evgenius:
We noticed a significant lag in the UI of our application.
We discovered that the "for" tag is not optimized for big contexts and when there's a lot of loops in the template.
The reason is the context being copied in every loop's body.
I.e. if you have 10 loops for an array of size 100, your context is copied for 1000 times.
It can take much time if the context is big.

evgenius:
https://github.com/justjohn/twig.js/commit/0282845a0cd41972f710bcffdfd89cb933205edd
Looks like we never merged it to the master :(
Anyway, going to change the way it works to avoid copying the context.

fourpixels:
What's up with this one? Don't see it in master, and I find it very important!

evgenius:
Hi @fourpixels, we optimized performance in 0.8.2. It's released already.

PR #211, opened 2015-02-16
"For" tag optimization

### LINKED ISSUE #93: Variable bug inside 2 loops
Hi,

There is a problem with variables inside 2 loops. I don't understand what is happening, but the variable get messed up.

Example:
http://pastebin.com/XADwCV3j

The same in JS:
http://pastebin.com/wXPjhLy7

evgenius:
Fixes #93, #144, #191, #210

### DIFF
--- a/src/twig.logic.js
+++ b/src/twig.logic.js
@@ -242,6 +242,7 @@ var Twig = (function (Twig) {
             parse: function (token, context, continue_chain) {
                 // Parse expression
                 var result = Twig.expression.parse.apply(this, [token.expression, context]),
+                    parent = context,
                     output = [],
 					len,
 					index = 0,
@@ -258,11 +259,11 @@ var Twig = (function (Twig) {
                             first: (index === 0),
                             last: isConditional?undefined:(index === len-1),
                             length: isConditional?undefined:len,
-                            parent: context
+                            parent: parent
                         };
                     },
                     loop = function(key, value) {
-                        var inner_context = Twig.lib.copy(context);
+                        var inner_context = context;
 
                         inner_context[token.value_var] = value;
                         if (token.key_var) {
@@ -279,6 +280,12 @@ var Twig = (function (Twig) {
                             index += 1;
                         }
                     };
+
+                // Store the parent as a prototype of the context
+                // It's probably the easiest and most natural way to split scopes
+                function Context() {}
+                Context.prototype = parent;
+                context = new Context();
 
                 if (result instanceof Array) {
                     len = result.length;
@@ -301,6 +308,18 @@ var Twig = (function (Twig) {
                         loop(key,  result[key]);
                     });
                 }
+
+                // Delete loop-related variables from the context
+                delete context['loop'];
+                delete context[token.value_var];
+                delete context[token.key_var];
+
+                // Merge changed context properties back to the parent
+                Twig.forEach(Object.keys(context), function (key) {
+                    if (key in parent) {
+                        parent[key] = context[key];
+                    }
+                });
 
                 // Only allow else statements if no output was generated
                 continue_chain = (output.length === 0);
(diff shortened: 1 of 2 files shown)


evgenius:
@jmealo Since you're active: mind looking into this?

jmealo:
@evgenius I don't think that I have time right now, but this definitely needs to get addressed.

evgenius:
@jmealo I might be not clear. It's a PR, not an issue to resolve.
But if you realize it's a PR and just don't have time to review it, no problem :)

evgenius:
Hi @justjohn ,

Could you please review this PR?
I'm too excited about using prototypes here so I may be missing something.

justjohn:
This looks solid. Using the prototype to extend the context seems like exactly the right approach, since it should work down multiple levels correctly. It may be worth defining a `Twig.Context` model more globally in the project, so this can be reused elsewhere.

evgenius:
@justjohn Thanks for reviewing it.
Defining Context makes sense only in that function scope. Otherwise we won't be able to change its prototype.
The idea is that you define a new Context function in the loop, set old context as its prototype (i.e. context parent) and instantiate a new context variable. If we go deeper (to a nested loop), a new function Context is declared and now the context above become a prototype of this new function.
So we can't define a global Context function.

justjohn:
This is more then should happen in this PR, but what I had in mind when I suggested a global Twig.Context object was mostly just encapsulating exactly what you wrote in a more reusable place. Something roughly like this:

```js
// Context object definition
Twig.Context = function(context) {
    // copy everything from context to this
    if (context) for (key in context) {
        if (context.hasOwnProperty(key)) {
            // prevent obscuring methods on base prototype
            if (!this[key]) this[key] = context[key]
        }
    }
}
Twig.Context.prototype = {
    _createChild: function() {
        function Context() {}
        Context.prototype = this;
        return new Context();
    }
}
```

Then to create the base context for the template you use:

```js
var context = new Twig.Context(user_provided_context);
```

And anytime you need to generate a child context in a tag, you can do:

```js
parse: function (token, context, continue_chain) {
    ...
    loop = function(key, value) {
        var inner_context = context._createChild();
    ...
}
```

Since `Twig.Context` lives at the top of the prototype chain for all contexts, `_createChild` is always available. This is a more reusable way of generating child contexts, since it may be necessary in more then just the `for` tag.

evgenius:
@justjohn I'm lost for words... Looks great :)
Let's merge this one and then implement your idea. Do you want to do it yourself or want me to create a commit? I'm not going to put my name as an author of that commit as it looks too awesome :)

justjohn:
I'll implement it and commit it this evening, since I've already written most of it :smiley:

evgenius:
@justjohn :
Isn't it more compact:
```js
Twig.Context = function (context) {
    function Context() {}
    Context.prototype = context;
    return new Context();
}

var context = Twig.Context(user_provided_context);
var inner_context = Twig.Context(context);
```

Am I missing something?

evgenius:
And we wouldn't need to have a service method `_createChild()` in the context object.

justjohn:
It is more compact, but it loses the benefits of being an object you can add service methods to. I see that as a benefit since it means they follow the context around. For example, if you're extending twig with your own custom for loop, you don't need access to the twig internals to create a child context.

justjohn:
I've added a `Twig.Context` object in ad55f6b59ab585d0f698bb06ea2fa83f156c75ed, It's pretty close to what's described here. I've also added a `_merge` function for folding changes from a child context back in, so all the logic now lives in one place.

evgenius:
@justjohn Great job, thanks.
Do you think it's time to release 0.8.0?

justjohn:
Yes, I think it is. Unless you have anything else you specifically want to get in, I'll push 0.8.0 up this evening.

evgenius:
Would be great, thanks!

createvibe:
Thank you guys so much for working on this!  It was much needed.  Can't wait for the release.

Issue #213, opened 2015-02-20
Issue using set while extending a template


ismyhc:
After upgrading from version 0.7.2 to 0.8 we are seeing a major issue using set while extending a template. Here is an example:

layout.html.twig
```
{% block body %}
	<div id="{{ pageId }}"></div>
{% endblock body %}
...
```

extended-layout.html.twig
```
{% extends 'layout' %}
{% set pageId = 'profile' %}
...
```

This worked as of 0.7.2.

evgenius:
Thanks Davis,

Looks like this case wasn't covered with tests.
I'll try to fix it asap.

evgenius:
@ismyhc, you can now upgrade to 0.8.1.

PR #214, opened 2015-02-20
Fix an issue with the "set" tag while extending a template

evgenius:


evgenius:
@justjohn I think we should revisit Twig.Context again as I have some concerns.
This PR fixes #213 though.

justjohn:
Merged and deployed as `0.8.1`

PR #215, opened 2015-02-22
Invert the intial context merge in Twig.Context

justjohn:
@evgenius This should address the performance issue with copying an initial context into a `new Twig.Context`. I've inverted the direction of the merge, so now the utility functions are merged into the existing context object.

I tested performance using this test (I didn't commit it, but you can drop it in test/test.perf.js and run with `./node_modules/.bin/mocha --require should test/test.perf.js`):
```js
var Twig = Twig || require("../twig"),
    twig = twig || Twig.twig;

describe("Twig.js Performance ->", function() {
    it("should output n values", function() {
        var ctx = {},
            n = 100000;

        for (var i=0; i<n; i++) {
            ctx['key'+i] = {'value': ctx['key'+(i-1)]};
        }

        var test_template = twig({data: '{% for value in test %}{{ value }}{% endfor %}'});
        test_template.render(ctx);
    });
});
```

The results were on the order:
Before `Twig.Context`: ~100ms
After `Twig.Context`: ~200ms
Updated `Twig.Context`: ~100ms

### DIFF
--- a/src/twig.core.js
+++ b/src/twig.core.js
@@ -118,6 +118,21 @@ var Twig = (function (Twig) {
         // 8. return undefined
     };
 
+    Twig.merge = function(target, source, onlyChanged, exclude) {
+        Twig.forEach(Object.keys(source), function (key) {
+            if (onlyChanged && !(key in target)) {
+                return;
+            }
+
+            // prevent context methods from being overwitten
+            if (!exclude || exclude.indexOf(key) === -1) {
+                target[key] = source[key];
+            }
+        });
+
+        return target;
+    };
+
     /**
      * Exception thrown by twig.js.
      */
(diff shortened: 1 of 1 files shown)


evgenius:
@justjohn , thanks, I'll take a look soon.

src/twig.core.js
```diff
@@ -166,20 +181,20 @@ var Twig = (function (Twig) {
...
     Twig.Context = function(context) {
+        context = context || {};
+
         // initializing Twig.Context over an existing context 
         // should return that context.
-        if (context instanceof Twig.Context) {
-            return context;
+        if (!context._twig_merge) {
+            for (var key in Twig.Context.Utils) {
+                context[key] = Twig.Context.Utils[key].bind(context);
```
evgenius: Hi @justjohn, I'm not sure it's okay to copy methods from `Twig.Context.Utils` into the context directly.
Now `{{ _context|keys }}` returns '_twig_create,_twig_merge'. Of course we can hack that filter (as well as others) but I think there should be a nicer way.

evgenius:
Honestly I'd like to get back to the initial idea:
1. Passed context doesn't change at all.
2. When we need a wrapped context to pass to the loop, we set the parent context as the prototype of the new context.
3. We have internal methods like `Twig.merge` and `Twig.Context` so we don't need to change the context adding all these `_twig_create()` and `_twig_merge()`.

I remember you wanted to have those methods so someone else can use it not digging into Twig internals, but to me modifying the initial context looks like a hack.
What do you think?

evgenius:
@justjohn That's how I did in our repository:
https://github.com/FoxyCart/twig.js/commit/8d41848f236a2a31b8899390a871c0187c3a4b9b

justjohn:
Ok, I'll pass on my goal of having the context methods passed around with the context. I'll merge in what you have in your repo and push a new version out with that.

PR #218, opened 2015-03-13
Fix for ie7

nike-17:
IE 7 throw an exception because of it in some cases.


evgenius:
Great catch, thanks @nike-17 !

evgenius:
@nike-17 Does it mean Twig.js works fine in IE7 (except this issue)?

nike-17:
We trying to provide core functionality in our web site for IE7 users - it looks like it more or less works with this fix

PR #228, opened 2015-04-03
extendTag can redefine an existing tag

xavierpriour:
`extendFunction` and `extendFilter` let you redefine existing functions and filters, but `extendTag` prevented it, throwing an error.

This patch changes it, so that calling `extendTag` with the same tag type will just replace the handling function.


xavierpriour:
@justjohn @evgenius just rebased to current code - does it need anything done before it can be merged? Thanks.

evgenius:
@xavierpriour Thanks for creating this PR. I had to clean it a bit: https://github.com/justjohn/twig.js/commit/6ea24766e9b141b3405c592c660af16f60a3dcb2
Of course your authorship is saved.

Currently I'm going to leave it in master.
We'll release a new version after we have some more changes.

Issue #237, opened 2015-06-16
property '_keys' in objects


plepe:
I just needed some custom function which expects an object as input, that there's a `_keys` property added to the object which holds the list of keys in the object, e.g.
```js
Twig.extendFunction("test", function(param) {
    return(JSON.stringify(param));
});
var t = twig({ data: "{{ test({ "foo": 1, "bar": 2 }) }}" });
t.render();
```

This will print:
```json
{"bar":2,"_keys":["foo","bar"],"foo":1}
```
See: http://jsfiddle.net/fzNNT/15/

(There you can also notice issue #82: wrong order of keys)

Twig functions/filters like `json_encode` remove this `_keys` property, but they do not do it recursively:

```
INPUT: {{ { "foo": 1, "bar": {"a": 1} }|json_encode }}
OUTPUT: {"bar":2,"_keys":["foo","bar"],"foo":1}
```
See: http://jsfiddle.net/fzNNT/16/

I think the only way forward is to remove the `_keys` property. I would assume that it is not really necessary, as it is not added to objects from render parameters:
```js
var t = twig({ data: '{{ val|json_encode }}' });
document.getElementById("result").innerHTML = t.render({ val: {"foo": 1, bar: { "a": 1 }}});
```
```
OUTPUT: {"foo":1,"bar":{"a":1}}
```
See: http://jsfiddle.net/fzNNT/17/

plepe:
After some thinking I decided it's cool to have the `_keys` property, as it's the only way to guarantee the correct order of object keys. Better fix filters which have problems with the `_keys` property and/or the order of keys like json_encode, which I did in pull request #279 .

Issue #82, opened 2013-04-29
merge|join outputs in the wrong order


m-abs:
From Twig's test/Twig/Tests/Fixtures/filters/merge.test

Template:
```
{{ items|merge({'bar': 'foo'})|join }}
{{ items|merge({'bar': 'foo'})|keys|join }}
{{ {'bar': 'foo'}|merge(items)|join }}
{{ {'bar': 'foo'}|merge(items)|keys|join }}
{{ numerics|merge([4, 5, 6])|join }}
```

Data:
```json
{ items: { foo: 'bar' }, numerics: [ 1, 2, 3 ] }
```

Expected result:
```
barfoo
foobar
foobar
barfoo
123456
```

Actual result:
```
foobar
barfoo
barfoo
foobar
123456
```

I think this is because the merge-filter uses "unshift" to add the keys to obj._keys and not push, because if I change it to push, the test is a success.

plepe:
No, I would assume that the problem is in the parser. Check the output of:
```
INPUT: {{ { "foo": 1, "bar": 2, "baz": 3 }|json_encode }}
OUTPUT: {"baz":3,"bar":2,"foo":1}
```

I've noticed this when I wrote a custom function, which needs an object as input. The order of keys was inverted.

dave-irvine:
I believe this was fixed in #279

PR #238, opened 2015-06-16
Test objects: retain key order; no additional information (like '_keys')

plepe:
Add tests for #82 and #237


plepe:
replaced by #279

PR #242, opened 2015-07-07
Implemented tests of array casting

torinaki:
Incorrect typecasting empty array to boolean. There are two cases:
1. Empty array not casted to `false`;
2. For some reason not empty array in `elseif` block cast to `false`;


dave-irvine:
I merged this in (kind of by accident in #343)

PR #246, opened 2015-07-14
Do not use `instanceof` on user input

roelvanduijnhoven:
See issue #156. Somewhere down the lines this fix got removed. Thus now when working with models that come from another frame, arrays are not detected correctly. This should fix that!


roelvanduijnhoven:
My bet, it was not removed. But it was not related to loops.

roelvanduijnhoven:
Just replaced all `instanceof` usages on Array and Object that originated from user input. This solves the problem completely.

Also rebased on recent master.

Could you review @evgenius because otherwise PR will get outdated again.

evgenius:
Thanks @roelvanduijnhoven !

PR #247, opened 2015-07-20
Update travis config to test on latest Node.js and on io.js

targos:


evgenius:
@targos Do you think we shouldn't test on 0.11?

targos:
@evgenius 0.11 was the development branch for 0.12. It is not maintained anymore, but you can keep testing against it if you think that people are still using it.

evgenius:
Ah, ok. Thanks for explaining. I'm fine with removing it.

Issue #243, opened 2015-07-10
Update sprintf.js in order to support "g" instruction


lpatiny:
Some improvement have been done recently on sprintf.js

One of them is to support "g": 
 case "g": arg = match[7] ? parseFloat(arg).toPrecision(match[7]) : parseFloat(arg); break

Would be nice to updated twig.js to the last published version

https://github.com/alexei/sprintf.js/blob/master/src/sprintf.js

PR #248, opened 2015-07-20
Update sprintf to v1.0.3

targos:
I copy-pasted the source from https://github.com/alexei/sprintf.js/blob/master/src/sprintf.js
and changed https://github.com/alexei/sprintf.js/blob/master/src/sprintf.js#L188-L207 to just return the functions that we need


targos:
this should fix #243

evgenius:
Great, thanks @targos .

PR #252, opened 2015-07-24
Normalize provided template path

stefanullinger:
Hi,

I've ran into a cross-platform issue that I think should be fixed within twig.js itself, so the behaviour is consistent across Unix and Windows systems.

In [grunt-twig-render](https://github.com/stefanullinger/grunt-twig-render) I am passing a template path to twig like this:

```
twig({
  path: template,
  async: false
});
```

The `template` variable is a unix file path like `some/template/path.twig` provided by the Gruntfile.
The template file includes another file (relative path) as follows:

```
{% include "partials/some-other-file.twig" %}
```

The issue now is that twig.js tries to resolve the relative path by splitting up the template path, removing the filename, and concatenating the relative file path. (see https://github.com/justjohn/twig.js/blob/master/src/twig.core.js#L1080 and following)

```
...
var base,
  base_path,
  sep_chr = "/",
  new_path = [],
  val;
...

if (template.url) {
  ...
} else if (template.path) {
  // Get the system-specific path separator
  var path = require("path"),
    sep = path.sep || sep_chr,
    relative = new RegExp("^\\.{1,2}" + sep.replace("\\", "\\\\"));
    file = file.replace(/\//g, sep);

  if (template.base !== undefined && file.match(relative) == null) {
    ...
  } else {
    base = template.path; // template.path = "some/template/path.twig", even on Windows
  }

  base = base.replace(sep+sep, sep);
  sep_chr = sep;
} else {
  ...
}

// sep_chr is system-specific, on Windows "\",
// so the following split operation does not split the string
base_path = base.split(sep_chr);
...

// on Windows, base_path is now: [ "some/template/path.twig" ]
// on Unix systems, base_path is now [ "some", "template", "path.twig" ]

```

This behavior can be consolidated by normalizing `template.path`. I am trying to do this at the topmost position in the code, so other or future functions can make use of this fix as well.

If you need some code to reproduce this behaviour, please check out [grunt-twig-render](https://github.com/stefanullinger/grunt-twig-render), run `npm install` and then start the tests via `grunt`. You will see that the tests are green on Unix systems but red on windows systems due to the inconsistant behavior described above.

Thanks for all your work on Twig.js!

Kind regards
Stefan


plepe:
I wanted to merge your branch into #280, but it does not work in browsers, as require() is only available to node.js.

stefanullinger:
Hi plepe, thanks for your review! I got your point and implemented the path normalization at another position, where we know, that it will only be used in Node.

plepe:
Looks reasonable. As there's actually only a single line which will be changed by this pull request, I'd rather prefer if you could compress it to a single commit. You can push to your repository with `git push -f` to force a new HEAD. You could also rebase your pull request on the current master.

Btw, here are the recently updated [Contribution guidelines](https://github.com/justjohn/twig.js/wiki/Contributing).

PR #267, opened 2015-10-16
Default filter now takes no params

dave-irvine:
The PHP version of twig allows you to use the default filter without passing any parameters.

See: https://github.com/fabpot/Twig/blob/master/lib/Twig/Node/Expression/Filter/Default.php#L29

An example would be:

```
{{ var|default|raw }}
```


plepe:
Why did you close this pull request? It has not been merged (to master) yet ...

plepe:
Ah, now I get it ... There was some mix-up with another pull request? The original pull request were only commits 4991873 and 393be7b?

PR #268, opened 2015-10-29
Twig.parse did not always return a string

christiaan:
Because parse directly returns Twig.output which can return either a String
or a string you never know which of the two you will get.

I did try to write a test for it but I was unable to figure out how to do it properly with should.js

This is the code we use in our project in which we work around this problem

```js
class TwigService {

  constructor(twigValues) {
    this._twigValues = twigValues;
  }

  /**
   * @param id The identifier for the template
   * @param html The twig template to parse
   * @param values Values that are used in the twig template
   * @return string
   */
  render(id, html, values) {
    var twigTemplate = Twig.twig({ ref: id });
    if (twigTemplate === null) {
      twigTemplate = Twig.twig({
        id: id,
        data: html,
        autoescape: true,
        rethrow: true
      });
    }

    values = merge.recursive(true, values, this._twigValues);
    values.date = Math.round(Date.now() / 1000);

    // Render returns a String object instead of regular string
    // casting it explicitly to a string fixes this odd behaviour
    try {
      return '' + twigTemplate.render(values);
    } catch(e) {
      return 'An error occured while rendering: ' + e;
    }
  }
}
```

### DIFF
--- a/src/twig.core.js
+++ b/src/twig.core.js
@@ -593,7 +593,7 @@ var Twig = (function (Twig) {
                         break;
                 }
             });
-            return Twig.output.apply(this, [output]);
+            return '' + Twig.output.apply(this, [output]);
         } catch (ex) {
             Twig.log.error("Error parsing twig template " + this.id + ": ");
             if (ex.stack) {


plepe:
I tried to merge your pull request, but it fails a test:
```
  1) Twig.js Functions -> Built-in Functions -> block -> shouldn't escape the content of blocks twice:

      AssertionError: expected 'te&amp;amp;st te&amp;amp;st' to be 'te&amp;st te&amp;st'
      + expected - actual

      +"te&amp;st te&amp;st"
      -"te&amp;amp;st te&amp;amp;st"
```

Could you please have a look at it again?

evgenius:
@christiaan 
A String constructor is used when the contents should be marked as valid Twig markup. It's how we detect if the contents were escaped already. If you convert a String object to a string in the `render()` method, it impacts the way inheritance works so you'll see double escaping. It's what @plepe mentioned above.
If a type of the output is important to you, you can use `template.render().toString()` to get a string.

christiaan:
I've looked at what our real problem was. When you pass a String to jQuery's .html() method it behaves differently then when it is passed a normal string.

It has cost us a couple of hours looking with 2 people to figure out what was going on. Especially since this behavior is not documented anywhere.

I'll see if I can get the tests running so I can come up with a different solution.

PR #270, opened 2015-11-06
Support `undefined` and `null` values for the `in` operator

nesk:
This code was failing:

```js
twig({
    data: '{{ 0 in undefined }} {{ 0 in null }}'
}).render();
```

Note that I wanted to add tests for this but I didn't found how to handle exceptions in your tests, this code never catches the exception:

```js
(function() {
    twig({
        data: '{{ 0 in undefined }} {{ 0 in null }}'
    }).render();
}).should.not.throw(TypeError);
```

Tried with the `expect` API but it doesn't work… You can either provide me instructions on how I should write those tests, or do it by yourself.

### DIFF
--- a/src/twig.expression.operator.js
+++ b/src/twig.expression.operator.js
@@ -17,10 +17,11 @@ var Twig = (function (Twig) {
     };
 
     var containment = function(a, b) {
-        if (b.indexOf !== undefined) {
+        if (b === undefined || b === null) {
+            return false;
+        } else if (b.indexOf !== undefined) {
             // String
             return a === b || a !== '' && b.indexOf(a) > -1;
-
         } else {
             var el;
             for (el in b) {


plepe:
Thanks for the pull request. I'm currently trying to merge some of the pull requests into a combined pull request. It would be great if you could create a test case which fails without your change.

nesk:
As I said, all the tests I tried to add were failing:

> Note that I wanted to add tests for this but I didn't found how to handle exceptions in your tests, this code never catches the exception:
> 
> ```js
> (function() {
>     twig({
>         data: '{{ 0 in undefined }} {{ 0 in null }}'
>     }).render();
> }).should.not.throw(TypeError);
> ```
> 
> Tried with the `expect` API but it doesn't work… You can either provide me instructions on how I should write those tests, or do it by yourself.

I need your help here :)

plepe:
Sorry, I didn't read your message too closely, or rather I read it but later-on forgot about it.

Why trying to catch the Exception? You could just test for the correct result. How about:
```js
        it("should support undefined and null for the in operator", function() {
            var test_template = twig({data: '{{ 0 in undefined }} {{ 0 in null }}'});
            test_template.render().should.equal('false false');
        });
```

On the other hand, although the above test case works for me, I'm not sure if this is the correct result. In Twig.php `{{ 0 in undefined }}` returns `null`:
`{{ 0 in undefined }}` -> (empty string)
`{% if 0 in undefined == null %}test{% endif %}` -> `test`

Therefore, The correct answer would be `.should.equal('');`

dave-irvine:
@plepe Can this be closed now? Think it landed in https://github.com/justjohn/twig.js/commit/65768548a10692e364ed8083bea74aeab0413498

plepe:
Sure. Apparently, I forgot about it.

PR #282, opened 2015-12-16
Filter 'sort': the compare function should return a number, not true/false

plepe:
On Safari the test for sorting objects failed. The result of the compare function should be <0 if a<b, 0 if a=b, >0 if a>b. The old compare function returned true/false.


plepe:
wrong base branch

PR #295, opened 2016-01-08
 Allow default filter to take 0 parameters

dave-irvine:
The PHP version of Twig allows default to be called without passing
any parameters, for example "{{ var|default|raw }}", with default
returning an empty string in this use case.

https://github.com/fabpot/Twig/blob/master/lib/Twig/Node/Expression/Filter/Default.php#L29

### DIFF
--- a/src/twig.filters.js
+++ b/src/twig.filters.js
@@ -172,10 +172,14 @@ var Twig = (function (Twig) {
             return output.join(join_str);
         },
         "default": function(value, params) {
-            if (params === undefined || params.length !== 1) {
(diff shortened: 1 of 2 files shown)


plepe:
Thanks!

Could you please remove the second commit from your pull request. twig.js will be rebuilt when releasing a new version. Also see [Contributing guidelines](https://github.com/justjohn/twig.js/wiki/Contributing).

dave-irvine:
@plepe Done

plepe:
Thanks!

PR #297, opened 2016-01-13
Added a demo for a compilation workflow.

hiwye:
I was fiddling with my template compilation workflow, so I decided to add this test html page, for new users to better understand a good templating workflow.

### DIFF
--- a/demos/compile_demo/compile_templates.js
+++ b/demos/compile_demo/compile_templates.js
@@ -1,0 +1,6 @@
+var exec = require('child_process').exec;
+var cmd = 'twigjs --output compiled/ --pattern *.twig raw/';
+
+exec(cmd, function(error, stdout, stderr) {
+    console.log(stdout);
+});
--- a/demos/compile_demo/README.md
+++ b/demos/compile_demo/README.md
@@ -1,0 +1,20 @@
+#Twig.JS Compilation Demo
+by Sebastian Walker
+##The workflow
+**Step 1**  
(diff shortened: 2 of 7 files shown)


plepe:
Sounds useful :) Only thing that I'm wondering if there should be a mechanism that the compiled versions get automatically updated when a new twig release is created (e.g. include it in the Makefile)? I guess that the compiled versions won't change much from twig release to twig release, but you can never be sure, can you?

hiwye:
Will work on it later. Thank you for the ideas!

-Sebastian

plepe:
Still working on this pull request?

hiwye:
I am, yes.
Have had a lot of client work to do recently, so not a lot of time for "open-sourcy" work.
I am hoping to get it done this weekend.

demos/compile_demo/compile_templates.js
```diff
@@ -0,0 +1,6 @@
+var exec = require('child_process').exec;
+var cmd = 'twigjs --output compiled/ --pattern *.twig raw/';
+
+exec(cmd, function(error, stdout, stderr) {
```
RobLoach: Since this is JavaScript, you could probably accomplish this without needing to invoke a child process :wink:

willrowe:
I'm going to close this since it is so old. @hiwye if you would like to submit another PR with everything updated to reflect the latest release, I'll be glad to review it.

Issue #294, opened 2016-01-08
[Request] override template loading mechanism(s)


r3wt:
allow replacement of xhr/template loading function.

This would be beneficial to my project, and maybe others as well.

plepe:
To pass the template as "data" parameter is not an option? Or do you need inline includes?

r3wt:
I'm already passing templates as data parameters, but it would be nice to have the xhr implementation settable so that i can take advantage of inline includes as well.

plepe:
I'd suggest you start hacking on it yourself. You can find the call to XMLHttpRequest in `src/twig.core.js`, line 936ff. I'm looking forward to a pull request :-)

r3wt:
10-4 plepe, i'll give it a go!

r3wt:
@plepe  where is Twig's constructor? i'm really having trouble understanding the pattern of this library.

plepe:
I haven't looked into Twig's internals yet either - better ask @justjohn , the founder. But the Twig constructor is in src/twig.core.js, right on top (there's not too much there, though).

r3wt:
@plepe How long do you estimate until this can land? would it be possible to have in next release?

r3wt:
why is this closed? the problem was not actually solved, since its not possible to load template via `{% include %}` in browser! ajax cannot be synchronous.... this really gets me worked up because its such an obviously necessary feature to even make this usuable in the browser. as it is, we are using it in one of our mobile apps with phonegap but the limitation is that we can't use inline includes, which leads to lots of extra code that belongs in subtemplates

plepe:
Sorry, I thought, that #301 solved your issue?

r3wt:
it allowed me to implement a custom loader sure, but there is still the issue of using inline templates. since twig.js apparently can't support asynchronous template loading without losing state, which ultimately made the solution worthless since i gain nothing. i had already implemented a 3rd party loading mechanism for my templates and then render with twig from string.... ultimately this does not provide a way to use inline templates in the browser, which is my only single gripe with twig.js everything else is great and performance is good enough to barely even notice an impact even with large complex templates.

plepe:
Hmm, I see. I created a pull request for tests concerning async inline include loading: #326 .

[...]

Issue #61, opened 2013-02-01
Promises with functions/filters?


ericclemmons:
Perhaps this is the wrong place to put this, but I was curious about soliciting input on this request...

Suppose I have a route like `/:page` that will simply `res.render(page + '.twig')`.  These views utilized lots of API information from services like Github, Basecamp, etc.

Suppose I have in my templates:

``` twig
{{ github_api('total_issues') }}
```

Would it be possible, if `github_api` return s a Promise, for the rendering engine to continue compilation & processing once the callback is called?

When our designers & developers write templates with Twig (PHP), the synchronous nature allows lazy-loading of resources without the controllers having to expose dozens if not hundreds of data-points that are mostly not used.

With Twig.js, the asynchronous nature means that the immediate return value is only what's available, and these data points would have to be pre-fetched, which would mean hundreds of unnecessary API calls.

I'm willing to code this functionality in, but as evident in my last ticket & PR, I'd need someone to point me to the pertinent sections to adjust (otherwise I'll be floundering).

Thanks!

Gertt:
Interested in this myself. We hacked together a super ugly workaround for an internal project at work, but it would be nice to have a way to let functions/filters preform async tasks.

I know it was mentioned in https://github.com/twigjs/twig.js/issues/357 that this is not available in PHP's Twig and therefor less important, but I'd also like to point out that PHP lets you do everything (web requests etcà synchronous while JS does not.

I might take a stab at providing a PR for this, but would like to know the maintainers' opinions on how best to implement this before starting out. Especially in regards to Promises vs callbacks etc. since this should also run in the browser and native promises are not supported yet everywhere. (I'd prefer Promise myself, personally though.)

dave-irvine:
If it was going to be done with anything I would suggest Promises. As to how to achieve it I have absolutely no idea I'm afraid!

Gertt:
After some looking around, I think it's doable. But if you want to maintain backwards compatibility, you'll probably end up with a complete mess in the code. I suppose the behaviour where `template.render` returns a synchronously rendered result is a must?

dave-irvine:
I think we would need to support both sync and async rendering going forward.

Potentially we could split the twig.js package up so that the renderers are separate, and you would then

`npm install twig twig-async`

or

`npm install twig twig-sync`

Any thoughts?

Gertt:
I am still looking into this and how exactly everything in the library fits together. The biggest problem I think will be existing custom tags.

[...]

PR #301, opened 2016-01-26
Override template loading mechanism (Issue 294)

createvibe:
This is an attempt to fulfill Issue #294, [Request] override template loading mechanism(s).

This PR introduces a way for users to extend Twig and implement their own custom template loaders.  It also allows you to override previously defined loaders.

For example, load a template from redis:

```
var Twig = require('twig'),
	redis = require('redis'),
	client = redis.createClient();

Twig.extend(function(Twig) {
	
	Twig.Templates.registerLoader('redis', function(location, params, callback, error_callback) {

		var precompiled = params.precompiled;

		if (typeof callback !== 'function') {
			throw new Twig.Error('The params.load callback is required for the redis loader method');
		}

		if (!params.async) {
			throw new Twig.Error('The "redis" loader method must be used asynchronously.');
		}

		client.get(location, function(err, data) {
			if (err) {
				if (typeof error_callback === 'function') {
					error_callback(err);
				}
				return;
			}

			if (precompiled) {
				params.data = JSON.parse(data);
			} else {
				params.data = data;
			}

			var template = new Twig.Template(params);
			callback(template);
		});

		return true;

	});

});
```

Fetch a template using the custom redis loader:

```
twig({
	method: 'redis', 
	name: 'my_template_name',
	load: function(template) {
		var output = template.render({data: 'here'});
	}
});
```

The business logic for 'fs' and 'ajax' loaders have been moved into their own files, twig.loader.fs.js and twig.loader.ajax.js, respectively.

Once you set the method for a template, all extends or includes spawning from that template should follow the same method.

Just as "path" and "url" are special to AJAX and FS methods, so is "name" special to custom methods.  However, custom methods also respect "id" if there is no name.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-SRC = src/twig.header.js src/twig.core.js src/twig.fills.js src/twig.lib.js src/twig.logic.js src/twig.expression.js src/twig.expression.operator.js src/twig.filters.js src/twig.functions.js src/twig.tests.js src/twig.exports.js src/twig.compiler.js src/twig.module.js
+SRC = src/twig.header.js src/twig.core.js src/twig.loader.*.js src/twig.fills.js src/twig.lib.js src/twig.logic.js src/twig.expression.js src/twig.expression.operator.js src/twig.filters.js src/twig.functions.js src/twig.tests.js src/twig.exports.js src/twig.compiler.js src/twig.module.js
 
 TESTS = test/*.js
 TESTSEXT = test-ext/*.js
--- a/src/twig.core.js
+++ b/src/twig.core.js
@@ -1034,6 +1034,8 @@ var Twig = (function (Twig) {
             base = params.base,
             path = params.path,
             url = params.url,
+            name = params.name,
+            method = params.method,
             // parser options
             options = params.options;
 
@@ -1056,9 +1058,11 @@ var Twig = (function (Twig) {
         //
 
         this.id     = id;
+        this.method = method;
         this.base   = base;
         this.path   = path;
         this.url    = url;
+        this.name   = name;
         this.macros = macros;
         this.options = options;
 
@@ -1122,7 +1126,7 @@ var Twig = (function (Twig) {
                 url = parsePath(this, this.extend);
 
                 ext_template = Twig.Templates.loadRemote(url, {
-                    method: this.url?'ajax':'fs',
+                    method: this.getLoaderMethod(),
                     base: this.base,
                     async:  false,
                     id:     url,
@@ -1155,7 +1159,7 @@ var Twig = (function (Twig) {
             if (!sub_template) {
                 sub_template = Twig.Templates.loadRemote(url, {
                     id: file,
-                    method: 'fs',
+                    method: this.getLoaderMethod(),
                     async: false,
                     options: this.options
                 });
@@ -1174,7 +1178,7 @@ var Twig = (function (Twig) {
 
         // Load blocks from an external file
         sub_template = Twig.Templates.loadRemote(url, {
-            method: this.url?'ajax':'fs',
+            method: this.getLoaderMethod(),
             base: this.base,
             async: false,
             options: this.options,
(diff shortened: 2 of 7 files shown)


r3wt:
:+1:

r3wt:
@plepe how long do you estimate until this can land?

plepe:
Sorry, been busy the previous days. The code looks very nice, it also has tests. I'd be willing to merge these changes.

@r3wt Did you test the change? Does it do what you expected?

r3wt:
@plepe I have not tested the changes, however i will do so in the morning and report back with results. it looks like it should do what i want. i really appreciate the hardwork that @createvibe has put into this.

createvibe:
@r3wt I'm happy to help out.  Hopefully it works for your needs.  Let me know if you need me to fix anything.

r3wt:
@createvibe i can't seem to build this.

img

createvibe:
@r3wt Interesting, it's building fine for me on a runnable instance. I just had to `npm install` first.

screenshot 2016-01-31 14 30 37

Looking at the list of files you are showing, it doesn't seem like it's trying to build the loaders.  Are you sure you are in the correct branch on my fork?

```
git clone https://github.com/createvibe/twig.js.git 
cd twig.js
git checkout issue_294
npm i
make clean
make
make test
```

r3wt:
ah, i see. i didn't run `npm install`. amazing how people create projects and don't document how to compile it.

createvibe:
Building the lib is documented a little here - https://github.com/justjohn/twig.js/wiki/Contributing

r3wt:
@createvibe i built it and wrote up a quick example and this is what i get:

img

maybe i did something wrong, i dunno. i cloned directly from your repo

createvibe:
@r3wt did you `git checkout issue_294` or are you directly on master?

r3wt:
@createvibe yes i did, and it works as far as loading a template. now let me reference a template inline and see what happens.

createvibe:
@r3wt Great!  Path lookups are a little different, but if all the templates come from your custom loader, you can possibly manipulate the path before fetching the file, inside your loader.

r3wt:
i get this error: `TwigException: Cannot extend an inline template.`






http://jqmvc.openex.info/seed/views/index.twig

createvibe:
For that, you need the "allowInlineIncludes" parameter set to true.

You can do this directly in the loader, if you want to make sure it's available, or you can do it by passing it as a twig() param.

If you look here, you can see how I implemented the tests: https://github.com/createvibe/twig.js/blob/issue_294/test/test.loaders.js#L8

r3wt:
its kind of fubar. inline template function of twig doesn't supply a callback, so i have no idea how i'm supposed to return the template to twig.

r3wt:
inline template doesn't work. twig spits out `undefined` for the filename and tries to fetch `undefined`:

img

i have checked and the loader is not being called with any parameters in this situation. all the arguments are undefined.

createvibe:
Could you show me how you are defining the loader and how you are rendering the template?

I have a working example here, if you want to see it: https://plnkr.co/edit/5wGPDF1fD4v2i0x3m70S

createvibe:
In response to twig passing async:false for inline includes, that's just how it works currently.  There's another thread about it (issue #61), but basically, twig needs to render expressions and logic blocks synchronously.  

You can see the code here for importFile, where it passes async:false manually: https://github.com/justjohn/twig.js/blob/master/src/twig.core.js#L1168

importFile is what {% include %} uses: https://github.com/justjohn/twig.js/blob/master/src/twig.logic.js#L650

r3wt:
well synchronous XHR is deprecated and blocked in the browser.

createvibe:
Does the plunkr I linked above crash your browser?

createvibe:
I do see that it's deprecated on the main thread, but it doesn't crash anything for me (chrome, ff, safari).  I'm not sure what to do about that except try to fix the synchronous loading, which is not an easy thing...

plepe:
I already tried to make rendering asynchronous, but it is 1. a lot of work 2. totally breaks compatibility, because render() can't return anything (it could return a promise ...) but would rather take a callback as additional parameter.

I guess for the inline templates, this happens while compiling; i haven't looked at that code yet ...

plepe:
By the way, do you think this pull request is ready for merging anyway? (synchronous inline templates are a different issue than custom loaders)

createvibe:
I've actually been looking into being able to return promises from logic blocks / tokens, and there's just a little bit of work to make it happen.  Problems come from existing code and extending other blocks, etc.  I will submit some test when I have time.

I have noticed a bug in the existing code that expects 'url' to be defined when it never is.  I will commit a fix for that, but as for this feature is concerned, I think it's good to go.

Let me know if you notice anything I should update.

r3wt:
:-1:

Issue #293, opened 2016-01-06
support block shortcuts


olets:
As per Twig docs (http://twig.sensiolabs.org/doc/tags/extends.html#block-shortcuts),
```
{% block title page_title|title %}
```
can be used in place of

```
{% block title %}
    {{ page_title|title }}
{% endblock %}
```

This currently breaks twig.js with "TwigException: Unable to parse". Would be great to see this supported!

olets:
Awesome thanks @createvibe

PR #304, opened 2016-01-30
issue #293 - support block shortcuts

createvibe:
Thought I would take a stab at this.  Let me know if there's a better way to do it.

This PR introduces block shortcuts (issue #293), as explained here: http://twig.sensiolabs.org/doc/tags/extends.html#block-shortcuts

Usage example:
```
{% block title page_title|title %}
```

Should be the same as:
```
{% block title %}
    {{ page_title|title }}
{% endblock %}
```


plepe:
As long as no one complains I guess your pull request is fine. Merged.

Issue #281, opened 2015-12-16
Running tests in browser: arr.should.containEql is not a function


plepe:
I'm currently trying the twig.js unit tests in different browsers. Beside a few small issues which I'll try to tackle soon, there's one problem which I can't figure out.

The tests for function "random" uses `should.containEql` to check whether the result is in the list of possible results. Apparently this function (containEql) is not available, when running the tests inside a browser (because the do work in node.js). I couldn't find any information about this on the interwebs ...

There are two versions of the resulting error message:

Mozilla Firefox 42, Chromium 47, Opera 12 report:
```
random ->
    should return a random item from a traversable or array
    arr.should.containEql is not a function
    @http://HOST/PATH/twig.js/test/test.functions.js:285:1
```

Safari (Windows), Gnome Webbrowser, XXXTerm report:
```
random ->
should return a random item from a traversable or array ‣
undefined is not a function (evaluating 'arr.should.containEql(twig({data: '{{ random(arr) }}'}).render({arr: arr}))')
http://HOST/PATH/twig.js/test/test.functions.js:285:42
```

I think that's exactly the same error, but a different message.

I haven't tested IE yet ...

evgenius:
@plepe How do you run the tests in a browser? In particular how do you include `should.js`?

plepe:
I was unsure myself, I've never used mocha or should.js before. I just browsed to the test/ directory (via https of course), and voila ...

PR #305, opened 2016-01-30
issue #281 - fix should.js in browser

createvibe:
This PR fixes the browser tests for should.js (issue #281).

As per the README at https://github.com/shouldjs/should.js, you just need to include "should.js" in the browser to get it working.  That means the NPM module needs to be installed first, by running `npm i`

I don't think we need Chai anymore, but I did not want to remove the lib in test/lib/chai.js.  All tests are passing now without it.


plepe:
Oh, great! Now all tests are working in the browser. Thanks!

PR #309, opened 2016-02-12
Source function implementation

bkotos:
Hello,

This pull request implements twig's [source](http://twig.sensiolabs.org/doc/functions/source.html) function, utilizing existing logic from ``` Twig.Templates.loadRemote ```.

### DIFF
--- a/src/twig.core.js
+++ b/src/twig.core.js
@@ -855,6 +855,12 @@ var Twig = (function (Twig) {
         loaders: {},
 
         /**
+         * Registered template parsers - use Twig.Templates.registerParser to add supported parsers
+         * @type {Object}
+         */
+        parsers: {},
+
+        /**
          * Cached / loaded templates
          * @type {Object}
          */
@@ -938,6 +944,66 @@ var Twig = (function (Twig) {
     };
 
     /**
+     * Register a template parser
+     *
+     * @example
+     * Twig.extend(function(Twig) {
+     *    Twig.Templates.registerParser('custom_parser', function(params) {
+     *        // this template source can be accessed in params.data
+     *        var template = params.data
+     *
+     *        // ... custom process that modifies the template
+     *
+     *        // return the parsed template
+     *        return template;
+     *    });
+     * });
+     *
+     * @param {String} method_name The method this parser is intended for (twig, source)
+     * @param {Function} func The function to execute when parsing the template
+     * @param {Object|undefined} scope Optional scope parameter to bind func to
+     *
+     * @throws Twig.Error
+     *
+     * @return {void}
+     */
+    Twig.Templates.registerParser = function(method_name, func, scope) {
+        if (typeof func !== 'function') {
+            throw new Twig.Error('Unable to add parser for ' + method_name + ': Invalid function regerence given.');
+        }
+
+        if (scope) {
+            func = func.bind(scope);
+        }
+
+        this.parsers[method_name] = func;
+    };
+
+    /**
+     * Remove a registered parser
+     *
+     * @param {String} method_name The method name for the parser you wish to remove
+     *
+     * @return {void}
+     */
+    Twig.Templates.unRegisterParser = function(method_name) {
+        if (this.isRegisteredParser(method_name)) {
+            delete this.parsers[method_name];
+        }
+    };
+
+    /**
+     * See if a parser is registered by its method name
+     *
+     * @param {String} method_name The name of the parser you are looking for
+     *
+     * @return {boolean}
+     */
+    Twig.Templates.isRegisteredParser = function(method_name) {
+        return this.parsers.hasOwnProperty(method_name);
+    };
+
+    /**
      * Save a template object to the store.
      *
      * @param {Twig.Template} template   The twig.js template to store.
@@ -972,6 +1038,8 @@ var Twig = (function (Twig) {
      *                      Defaults to true.
      *      method:      What method should be used to load the template
      *                      (fs or ajax)
+     *      parser:      What method should be used to parse the template
+     *                      (twig or source)
      *      precompiled: Has the template already been compiled.
      *
      * @param {string} location  The remote URL to load as a template.
@@ -1004,9 +1072,12 @@ var Twig = (function (Twig) {
             return Twig.Templates.registry[params.id];
         }
 
+        //if the parser name hasn't been set, default it to twig
+        params.parser = params.parser || 'twig';
+
         // Assume 'fs' if the loader is not defined
         loader = this.loaders[params.method] || this.loaders.fs;
-        return loader.apply(null, arguments);
+        return loader.apply(this, arguments);
     };
 
     // Determine object type
--- a/src/twig.functions.js
+++ b/src/twig.functions.js
@@ -7,6 +7,11 @@
 //
 // This file handles parsing filters.
 var Twig = (function (Twig) {
+    /**
+     * @constant
+     * @type {string}
+     */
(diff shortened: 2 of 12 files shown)


plepe:
I just wanted to check your pull request, but there are conflicts when merging, because in the last version 0.8.8 the loadRemote() function has been redesigned in #301. Could you check if you could rebase your pull request on the current master? Thx.

src/twig.core.js
```diff
@@ -960,10 +968,18 @@ var Twig = (function (Twig) {
                         params.url = location;
                         params.data = data;
 
-                        template = new Twig.Template(params);
+                        if (fetchTemplateSource === true) {
+                            templateSource = data;
```
plepe: Why are you using a separate `templateSource` variable? If you just use `template`, you don't need to change the callbacks resp. return below ...

```js
                        if (fetchTemplateSource === true) {
                            template = data;
                        } else {
                            template = new Twig.Template(params);
                        }
```

src/twig.functions.js
```diff
@@ -232,6 +236,57 @@ var Twig = (function (Twig) {
...
+            var templateSource;
+            var templateFound = false;
+            var options = {
+                id: 'templates/source.twig',
+                method: 'ajax',
+                async: false,
+                fetchTemplateSource: true
+            };
+
+            //default ignore_missing to true
```
plepe: I think, in Twig.php ignore_missing defaults to false? Also, it throws an exception ...

src/twig.core.js
```diff
@@ -913,15 +916,20 @@ var Twig = (function (Twig) {
...
-        var id          = params.id,
-            method      = params.method,
-            async       = params.async,
-            precompiled = params.precompiled,
-            template    = null;
+        var id                  = params.id,
+            method              = params.method,
+            async               = params.async,
+            precompiled         = params.precompiled,
+            fetchTemplateSource = params.fetchTemplateSource,
```
plepe: How about calling the option `format` or `parser`? We could introduce a function to add available parsers, e.g. to include markdown files.
```js
parser: "twig"
parser: "source"
parser: "markdown"
```

Just my ideas ...

src/twig.core.js
bkotos: good point, I'll get rid of ``` templateSource ```

src/twig.core.js
bkotos: I like that idea.  That will make it easier to add different types of formats in the future

src/twig.functions.js
bkotos: good catch, I'll fix that

bkotos:
I'll be updating this PR within the next few days.  I will let you know when it's ready.

plepe:
How about progress? I think it's time to compile a new version ...

bkotos:
@plepe my apologies for the late response to this.  I've updated the PR to be compatible with the new loading system.  In doing this, I implemented abstract parsers like we had talked about.

@plepe and @createvibe let me know what you think.  I'd be glad to change anything that doesn't look right to you guys. :)

test/test.parsers.js
```diff
@@ -0,0 +1,36 @@
...
+        });
+
+        it("should run the data through the custom parser", function() {
+            Twig.extend(function(Twig) {
+                var params = {
+                    data: 'This is a test template.'
+                };
+                var template = Twig.templates.parsers.custom(params);
+
+                template.should.equal('[CUSTOM PARSER] This is a tes template.');
```
plepe: Typo: "This is a tes_t_ template"

plepe:
Wow, that looks pretty awesome.

Do you think you could tidy up the commit history a little bit? E.g. when you fix a typo rather update the old commit, instead of creating a new commit? You are changing a parameter name from `parser` to `parser_method` and later-on back to `parser`.

I think it's important to have distinct commits, so that you can follow development process, but readability suffers from too many commits ...

bkotos:
@plepe should be more readable now.  Squashed most of the commits.

test/test.parsers.js
bkotos: Fixed

test/test.functions.js
```diff
@@ -319,7 +319,7 @@ describe("Twig.js Functions ->", function() {
 
         describe("source ->", function() {
             it("should load the non-compiled template source code", function() {
-                twig({data: '{{ source("test/templates/source.twig") }}'})
+                twig({data: '{{ source("/test/templates/source.twig") }}'})
```
plepe: Should this really be an absolute path?

plepe:
I'm sorry, I'm still not totally happy with this history. The first commit looks fine on first look.

The second - "source function" - also changes things of the parser, e.g. obviously a mistake in the isRegisteredParser() function - also the 'parser_method' parameter gets renamed to 'parser'.

The third - "finally fix ... tests" not only changes tests as I would expect it should. In fact, it changes the loader and the source function.

Sorry. I really think that it's important to have a tidy history where you can follow development process.

plepe:
Btw, I think the structure of commits: 1) implement registerParser and similar 2) source function and so is fine.

bkotos:
Ok.  So just to be clear, you would like to see the last two commits https://github.com/justjohn/twig.js/pull/309/commits/207906e6f4c995301bb2845905c314a5b0459967 and https://github.com/justjohn/twig.js/pull/309/commits/f22f00367b2b733564723dbe814f0acfa0632ae1 change?

Also, would it be more sensible if I moved the changes from https://github.com/justjohn/twig.js/pull/309/commits/207906e6f4c995301bb2845905c314a5b0459967 into the first commit, that implements the source function?

test/test.functions.js
```diff
@@ -316,5 +316,31 @@ describe("Twig.js Functions ->", function() {
                 twig({data: '{{ max({2:"two", 1:"one", 3:"three", 5:"five", 4:"four"}) }}'}).render().should.equal('two');
             });
         });
+
+        describe("source ->", function() {
+            it("should load the non-compiled template source code", function() {
+                twig({data: '{{ source("/templates/source.twig") }}'})
```
dave-irvine: Shouldn't this be
```
twig({data: '{{ source("templates/source.twig") }}'})
```

i.e no leading "/" on "templates"?

plepe:
Now we are getting somewhere :) Very nice!

As you might have noticed, one of the unit tests failed - the one where @dave-irvine commented about the leading "/".

The only thing that I might have changed (maybe this is a bit nit-picky of me) to move the `src/twig.parser.source.js` file to the commit where you implement the source function.

bkotos:
Haha, I was just about to fix the unit test and ping you, but I can see you merged this already.  Anyhow, thanks for merging. :)

src/twig.core.js
```diff
@@ -1004,9 +1072,12 @@ var Twig = (function (Twig) {
...
         }
 
+        //if the parser name hasn't been set, default it to twig
+        params.parser = params.parser || 'twig';
+
         // Assume 'fs' if the loader is not defined
         loader = this.loaders[params.method] || this.loaders.fs;
-        return loader.apply(null, arguments);
+        return loader.apply(this, arguments);
     };
```
createvibe: I think we have to be careful with this line.  Since the loaders can be bound to a different scope during registration, applying to 'this' here, might not do what you expect.  It shouldn't affect the existing built-in loaders since they do not bind to a scope object, but if you look at the loader tests, it does.  In the tests, though, it does not take the parser into consideration, and so there are no problems.  But I thought I would point it out...

Issue #160, opened 2014-05-23
Defined test not working properly on object properties


targos:
Unlike the original version of twig, a condition like "{% if foo.bar is defined %}" is always true if foo exists.
The reason is that if "foo" is defined and it has no "bar" property, then the value "null" is sent to the "defined" test. Null is not considered undefined so the test always returns true...

Testcase : http://jsfiddle.net/QJA7z/1/

alexeymalihin:
Any updates on this? :up:

adrienrn:
Confirm this behavior for defined test on undefined property of objects :) !

plepe:
Confirmed. I just pushed tests for these issues, based on the commit from @jopdeklein. Unfortunately the referenced 'fix' does not work (see the second commit in the pull request #310).

plepe:
Yep, found it, should be fixed. In fact it was just one small change, that if a property of an object can't be found, it should use the value `undefined`, but it used the value `null`.

dave-irvine:
Believe this is resolved in #310.

PR #310, opened 2016-02-16
Fix "is defined" test

plepe:
see #160

### DIFF
--- a/src/twig.expression.js
+++ b/src/twig.expression.js
@@ -647,7 +647,7 @@ var Twig = (function (Twig) {
                 } else if (object["is"+capitalize(key)] !== undefined) {
                     value = object["is"+capitalize(key)];
                 } else {
-                    value = null;
+                    value = undefined;
                 }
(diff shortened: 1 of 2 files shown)


plepe:
Anyone (e.g. @createvibe because you are quite active recently) wants to review this pull request? See #160 for explanation.

createvibe:
@plepe I found an issue with nested objects.

```
var context = {
    key: {
        nested: {
            data: "nested data"
        }
    }
};

twig({data: '{{ key.nested.fail is defined }}'}).render(context).should.equal( "false" );
```

The above is returning "true" instead of "false" and causes the test to fail.

It is returning true on "key.nested.anything" because key.nested is defined.  I don't think it's looking further than the first period.

Is this expected?

createvibe:
shoot I'm sorry @plepe, I spoke too soon.  For some reason make didn't want to regenerate the twig files.  it all works good.  CR passes.  :)

plepe:
Awesome :-) Then I'll merge.

PR #320, opened 2016-03-08
updated to allow namespaces to be used with @

gravypower:
Updated twig.js to allow the use if @ when using namespaces.

I made this change because of the documentation found [here ](http://twig.sensiolabs.org/doc/api.html#built-in-loaders).

> Namespaced templates can be accessed via the special @namespace_name/template_path notation

### DIFF
--- a/src/twig.core.js
+++ b/src/twig.core.js
@@ -1270,11 +1270,14 @@ var Twig = (function (Twig) {
         if (typeof template === 'object' && typeof template.options === 'object') {
             namespaces = template.options.namespaces;
         }
-
-        if (typeof namespaces === 'object' && file.indexOf('::') > 0) {
+        
+        if (typeof namespaces === 'object' && (file.indexOf('::') > 0 || file.indexOf('@') >= 0)) {
             for (var k in namespaces){
                 if (namespaces.hasOwnProperty(k)) {
+
                     file = file.replace(k + '::', namespaces[k]);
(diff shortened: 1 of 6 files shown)


gravypower:
My first attempt with writing javascript code for a pull request and am not 100% sure I have created the tests correctly or even if the way I have implemented this is the best way.

plepe:
I haven't looked into it too detailed, it looks pretty reasonable. Could you please update the pull request to not include the changes in twig.min.js and similar files, as outlined in [the contribution guidelines](https://github.com/justjohn/twig.js/wiki/Contributing).

gravypower:
Updated, sorry I should have read them first.

plepe:
Waiting for discussion in #323 before merging.

plepe:
Sorry for pointing this out so late, but would you mind merging your two commits into a single one (`git rebase -i` helps a lot)? You just need to do a `git push --force` to update your pull request. If you want I can also do this as part of #327, but then your pull request wouldn't show up as merged.

gravypower:
Sure thing, I will give it a go.

gravypower:
OK I think I worked it out.

plepe:
Yeah, thanks. Sorry, but I think it was unnecessary - @dave-irvine just said, he'd like to re-implement your function on top of pull request #323.

gravypower:
All good, looks like it will be done in a more robust manner anyhow.

plepe:
Replaced by #328.

Issue #225, opened 2015-03-18
In the PHP version, you can include templates from multiple locations


nebulousGirl:
It would be nice to be able to pass an array of locations for the path option.

brunowego:
+1

mikelohmann:
+1

micahgodbolt:
+1

marsjaninzmarsa:
+1

fi5u:
+1

raynor85:
+1

rafaellyra:
if I understood you are talking about namespaces, it is now working at the last version: https://github.com/justjohn/twig.js/blob/master/CHANGELOG.md

and I added a example on how to implement on the [wiki](https://github.com/justjohn/twig.js/wiki#namespaces)

nebulousGirl:
Namespaces are a great addition, but I am talking about cascading locations.

In php, you can specify an array of folders where to look for templates.

Twig will loop through the locations until it finds the template.

If I define my path as [folder-a, folder-b], Twig looks for template.twig first in folder-a and if it doesn't exist, it will cascade to look if it exists in folder-b.

micahgodbolt:
in Express 4 we'd hope to set it up like

```
app.set('views',  ['path/to/files/', 'path/to/other/files'] ;
```

But when you do this, Twigjs doesn't see this as an array of 2 items. It tries to look in:

```
'path/to/files/,path/to/other/files'
```

micahgodbolt:
Appears that the `relativePath` function assumes template.base to be a string, and not an array (which it is when multiple paths are defined in express.

I took a little stab at a fix, but it might require more of a rewrite than I'm able to do in order to support array paths.

JohnAlbin:
> Namespaces are a great addition, but I am talking about cascading locations.
>
> In php, you can specify an array of folders where to look for templates.
>
> Twig will loop through the locations until it finds the template.

Actually, they are the same thing. In the PHP Twig, there is a default "main" namespace. When you add paths to search for templates, you are adding paths to the main namespace. Here's the relevant docs: http://twig.sensiolabs.org/doc/api.html#built-in-loaders

There are 2 incompatibilities with Twig.js here.

1. Twig.js does not have a main namespace.
2. Namespaces can only have a single path. But namespaces should have an ordered array of paths.

[...]

PR #321, opened 2016-03-10
Support arrays in express views / template.base

micahgodbolt:
Hopefully this solution is simple enough to be a drop in fix for supporting multiple template locations.

Since the relative path function assumes a single string path, the goal is to check for the existence of the file on all of the paths, and then return the path to the reset of the function if that file exists.

The order of paths are flipped so that if the file is located in more than 1 location, the location earlier in the array will win out.

All the current tests pass, but I'm not sure how to approach adding a test for this functionality.

### DIFF
--- a/src/twig.core.js
+++ b/src/twig.core.js
@@ -1300,21 +1300,20 @@ var Twig = (function (Twig) {
             val;
 
 
-	if (typeof template.base === "object") {
-	  var fs = require("fs"),
-	      path = require("path"),
-	      temp = template.base.reverse(),
-	      sep = path.sep || sep_chr;
+	if (Array.isArray(template.base)) {
+	    var fs = require("fs"),
+		path = require("path"),
+		temp = template.base.reverse(),
+		sep = path.sep || sep_chr;
 
-	  for (var key in temp){
-	    try {
-	      fs.accessSync(temp[key] + sep + file, fs.F_OK);
-	      template.base = temp[key];
-	    } catch (e) {
+	    for (var key in temp) {
+		try {
+		    fs.accessSync(temp[key] + sep + file, fs.F_OK);
+		    base = temp[key] + ((temp[key].charAt(temp[key].length-1) === '/') ? '' : '/');
+		} catch (e) {
+		}
 	    }
-	  }
-	}
-	if (template.url) {
+	} else if (template.url) {
             if (typeof template.base !== 'undefined') {
                 base = template.base + ((template.base.charAt(template.base.length-1) === '/') ? '' : '/');
             } else {
--- a/src/twig.core.js
+++ b/src/twig.core.js
@@ -1299,7 +1299,19 @@ var Twig = (function (Twig) {
             new_path = [],
             val;
 
-        if (template.url) {
+
+	if (typeof template.base === "object") {
+	  var fs = require("fs"),
+	      path = "";
+
+	  for (var key in template.base){
+	    path = template.base[key] + '/' + file;
+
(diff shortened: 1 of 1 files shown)


micahgodbolt:
Fixes #225

createvibe:
Well I was really just worried about this triggering when it wasn't intended to.  I ran the browser tests and it seems ok.  I don't think checking typeof === 'object' is safe though, considering new String('') and so many other things would make that evaluate to true.  Looking at your usage, it seems like you want an array, so why not check for instanceof Array?

Something else I noticed is that you are changing template.base with the matching path.  I understand that you don't want to check for the existence of the file more than you have to, but if something on the server creates the file after the template is rendered, the next time the template runs it won't find it because template.base would have been rewritten.  Normally this wouldn't be a problem, but if the template is served from cache, depending on the environment you are running this in, you may never end up with the right file until you restart the server or clear the template from cache, which I don't think is supported yet.  This may or may not pose a problem, depending on your application.  But please correct if I am wrong or missed something.

Also, with the adoption of different loader methods, it might be useful to check the loader method to make sure we are on 'fs' before checking that file paths exist.  So maybe adding a check for template.getLoaderMethod() === 'fs' would be helpful, or perhaps just move this code down inside the check for template.path.

One thing that bugs me a little, is that if template.url, twig assumes it's ajax, and if template.path, twig assumes it's fs.  Maybe this should be based on the loader instead of the property.  But if we do that, I'm wondering if the loader should provide this business logic itself.  Maybe the loaders can be redefined a little so that they can define methods like, relativePath.  If that were the case, we could say something like loader.getRelativePath(template, file) and the business logic could be moved there, appropriately.

@plepe what do you think?

micahgodbolt:
I at first thought array would be the correct choice, but a quick typeof test showed it was actually an object. Also, I can't imagine a situation where this is triggered when not intended. If a string is passed, nothing happens, if an object/array is passed the current implementation is completely broken.

So sure, there might be a better way to refactor the entire process to handle that value better, but without a total rewrite of this and several other functions, this approach at least solves the problem for now until a better solution is created.

micahgodbolt:
Drat. This works for the first include, but I didn't realize `template.base` was persistent for the entire rendering process. So once it is changed from object to string, it won't work for the next include.

So need to set it back to the original value, or use some temp value instead.

micahgodbolt:
I was able to get the array check to worry properly with Array.isArray, and found that updating template.base made subsequent function calls not work properly. So now the array check passes the correct path directly to `base` and uses same logic for appending a '/' as the logic right below it

```
base = temp[key] + ((temp[key].charAt(temp[key].length-1) === '/') ? '' : '/');
```

plepe:
Could you please add tests for loading from multiple locations?

micahgodbolt:
I'm not 100% sure how to simulate multiple template locations in this testing environment. I'm setting those paths in the express. I'm assuming you don't want express added to the dev dependencies.

plepe:
Hmm, I see. Could you show an example how you use the multiple locations? Is this code only available when using express?

Also, would you mind tidying up your commit history a little bit? I think it's important to have distinct commits, so that you can follow development process. In my impression you could squash your commits into a single or two commits. Also it might be helpful to use the commit message of your third commit as comment in the code (about reversing the array).

plepe:
Your pull request conflicts with the changes introduced in #323. Could you please merge the current master into your pull request or better rebase it?

micahgodbolt:
Closing this PR as #332 is a cleaner MR with master rebased back in.

PR #324, opened 2016-03-16
fix IE8 by using Twig.indexOf fallback

vvanghelue:
I got an error en IE8 using blocks. Problem fixed by using the Twig.indexOf created for this purpose.


dave-irvine:
I can already tell you @plepe will ask you to do another commit without the compiled files included. :)

plepe:
@dave-irvine That's right :) Please check the [Contribution guidelines](https://github.com/justjohn/twig.js/wiki/Contributing) for details. You know, you can update the pull request by using "git push -f" (force).

Apart from that, the changes look reasonable to me.

plepe:
Merged in 0bf1a76cf53cb784e1692f578171250b022f03d0

createvibe:
@plep can we document the browser versions that should be supported?  Or is it already and maybe I just missed it?

plepe:
I think there is no such documentation. Personally, I think we don't need to support IE8 any more, but Twig.indexOf and similar functions are used throughout the code, therefore I think it's okay to use it in this location too ...

PR #326, opened 2016-03-16
Browser tests for inline includes, using asynchronous loading

plepe:
Currently includes commits from #325, will update as soon as these are merged to master.

Currently async loading for inline includes is not supported.


dave-irvine:
@plepe If the tests are passing, does that mean the functionality exists? If so we can merge this.

plepe:
If I run the tests, they are not passing. I guess you tried only with nodejs? These tests test async loading within browser context.

dave-irvine:
Ok that means we need a test runner for browser tests. Zombie.js ?

plepe:
I don't know. I just run them within Firefox and other browsers on my desktop. To have some CI method would be advantageous.

willrowe:
I'm going to close this since it is so old. @plepe if you would like to submit another PR with everything updated to reflect the latest release, I'll be glad to review it.

PR #332, opened 2016-03-22
Caught up with master, merged in array support changes in single commit

micahgodbolt:
- Check for array of views and test for file existance before passing path
- Reversed arrary so that first path would replace second path if file were in both locations

Replaces #321 with a rebased branch with single commit.

### DIFF
--- a/src/twig.path.js
+++ b/src/twig.path.js
@@ -60,7 +60,21 @@ var Twig = (function (Twig) {
             file = file || "",
             val;
 
-        if (template.url) {
+
+	if (Array.isArray(template.base)) {
+		  var fs = require("fs"),
+			path = require("path"),
+			temp = template.base.reverse(),
+			sep = path.sep || sep_chr;
+
+		  for (var key in temp) {
+			try {
+			    fs.accessSync(temp[key] + sep + file, fs.F_OK);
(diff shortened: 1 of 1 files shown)


dave-irvine:
Hi @micahgodbolt 

Is there any chance of getting some tests for this functionality?

micahgodbolt:
I personally wouldn't know how to test this. For my usecase, the multiple paths are injected via Express.js. So without putting this into Express, I am not sure how to specify those paths.

dave-irvine:
If you rebase to latest master you could require this file specifically in your tests and write a unit test rather than an integration test. Basically checking that when `template.base` is an Array it calls your code, and that base then ends up how you expect it.

JohnAlbin:
IMO, this functionality should not be implemented as in this PR. Instead it should be implemented in Twig namespaces while ensuring that Twig.js has a main namespace that it uses by default. https://github.com/twigjs/twig.js/issues/225#issuecomment-217335847

I just added a Twig.js builder to https://github.com/kss-node/kss-node/ and want to improve Twig.js's namespace support.

/me looks at master.

Wow! you've done a lot of refactoring since 0.8.9. :-) It will take me a while to figure out the new code. I only just got my head wrapped around 0.8.9.

PR #336, opened 2016-03-24
Add typescript definitions

### LINKED ISSUE #333: TypeScript support
Typescript slowly moves away from the global definitions repository and each project maintains their own definitions. Would be great for see this project follow.

ianbytchek:
Closes #333 by providing a most common definitions for the module and adding `typings` field in `package.json` to enable automatic discovery by typescript compiler / ides.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "url": "git://github.com/justjohn/twig.js.git"
   },
   "main": "twig.min.js",
+  "typings": "twig.d.ts",
   "engines": {
     "node": "*"
   },
--- a/twig.d.ts
+++ b/twig.d.ts
@@ -1,0 +1,143 @@
+export interface Token {
+    type:string;
+
+    open?:any;
+    close?:any;
+    regex?:RegExp;
+    next?:any[];
+    output?:any;
+    match?:string[];
+
+    parse?(token:any, context:any, chain:any):ParsedToken;
+    compile?(token:any):CompiledToken;
+}
+
+export interface CompiledToken {
+    type:string;
+
+    value:{};
+    stack?:Token[];
+    output:CompiledToken[];
+}
+
+export interface ParsedToken {
+    chain:any;
+    output:string;
+}
+
+export interface TokenGroup {
+    type:any;
+    definitions:Token[];
+
+    extend(definition:Token):void;
+    compile(token:Token):CompiledToken;
+    tokenize(expression:string):Token[];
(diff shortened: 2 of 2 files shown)


dave-irvine:
I'd rather these go into http://definitelytyped.org/ instead.

ianbytchek:
This is against best practices – DefinitelyTyped project outgrew itself and now moving towards definitions being kept along with the project sources.

ianbytchek:
Besides, there's a huge benefit of keeping them along with the distribution, it automatically gets detected by compiler and most IDEs, removing the need manually managing d.ts dependencies.

dave-irvine:
Who's best practices? :) DefinitelyTyped seems to still be being regularly updated so I wouldn't call it abandoned and I can't see anything there that says not to submit any new ones.

My problem with adding them to the repo is that it is then something else that needs to be maintained by the core project. If they are in a seperate repo (I'm not particularly attached to *which* repo), then it no longer needs maintaining as part of the core project.

It looks like I'm the primary maintainer right now and I have no use for TS definitions so it would just complicate fixing bugs if I need to go in and maintain those definitions for every change. Right now I'd rather focus on fixing pure JS issues.

ianbytchek:
DefinitelyTyped is a huge project, it's maintained and will be for awhile, but the general direction has changed – the preferred solution is not to store everything in a single ginormous repository. I'm not an insider on the matter and might have misinterpreted some information, correct me if wrong, but overall it sounds like common sense:

- https://twitter.com/typescriptpdx/status/692195296546287616
- http://stackoverflow.com/questions/35598876/why-is-tsd-deprecated
- https://github.com/DefinitelyTyped/tsd/issue

That said, totally understand your reasons for not adding unnecessary fat to the project. I am actively using and planning to use twig.js in foreseeable future, with already maintaining local definitions I'd be ready to handle associated issues arising in this repo. Manually copying definitions from project to project is criminal, doing it with definitely typed is better approach, but adds extra steps with own problems, I'd rather do it properly.

dave-irvine:
OK I think Typings is the new state of the art? https://github.com/typings/typings

For example, for the bluebird npm package, typings has this in its repository:

https://github.com/typings/registry/blob/master/npm/bluebird.json

and bluebird doesn't seem to maintain its own TS definitions:

https://github.com/petkaantonov/bluebird

ianbytchek:
Yep, typings is the future, apparently. Ok then, let's do it the bluebird way.

PR #340, opened 2016-04-11
Replace Twig.exports.cache by setCache

AntoineViau:
As Twig.exports.cache is a function, this naming might be confusing for the user. setCache fits better.

### DIFF
--- a/twig.js
+++ b/twig.js
@@ -6514,7 +6514,7 @@ var Twig = (function (Twig) {
      *
      * @param {boolean} cache
      */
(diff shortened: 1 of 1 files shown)


dave-irvine:
Hi,

I think because this changes the public API this will need to wait for a 0.9 release.

willrowe:
I'm going to close this since it so old and out of sync with the repo. Feel free to open an issue if you feel this is still needed and it can be considered for the next major release.

PR #348, opened 2016-04-20
Return rendered template as a string (for express)

DracoBlue:
Because https://github.com/expressjs/express/blob/b5a2801/lib/response.js#L136 does not work with `String`, which is created at https://github.com/twigjs/twig.js/blob/274a3ce830be07f10a9d1c09d7a2f671adceb649/src/twig.core.js#L1319

If this patch is not applied and you use twig.js with express like this:

``` js
app.set('twig options', {
  autoescape: 'html'
});
```

then the returned value would be a "String" and not a "string". So express handles this by returning the value as `res.json()` :(.

So the html page looks like this:

``` json
"\n<!DOCTYPE HTML>\n<html la.."
``` 

Of course the HTML should not be delivered as JSON if the user wants the contents to be escaped ;).

PS: Workaround for others with the same issue:

``` js
app.engine('twig', function(path, options, fn) {
  Twig.__express(path, options, function(err, response) {
    fn(err, "" + response);
  });
});
```

### DIFF
--- a/src/twig.exports.js
+++ b/src/twig.exports.js
@@ -174,8 +174,8 @@ module.exports = function (Twig) {
             path: path,
             base: settings.views,
             load: function(template) {
-                // render and return template
-                fn(null, template.render(options));
+                // render and return template as a simple string, see https://github.com/twigjs/twig.js/pull/348 for more information
+                fn(null, '' + template.render(options));
             }
         };
 
--- a/test/test.exports.js
+++ b/test/test.exports.js
@@ -1,0 +1,21 @@
+var Twig = Twig || require("../twig"),
+    twig = twig || Twig.twig;
+
+describe("Twig.js Exports __express ->", function() {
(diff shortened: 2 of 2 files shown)


DracoBlue:
I would like to add tests for this. Where does one add tests for express?

dave-irvine:
Hi,

I don't think we've ever had any specific web-server tests before.

In this case I think go ahead and make `test.exports.js` and write your tests in there.

Also could you add a comment above the line you've added to explain why you concatenate with a `''`, maybe with the number of this issue (#348) so that it is obvious why you've done it in the future :)

Thanks!

DracoBlue:
@dave-irvine I cleaned up the commits and added a test.

PS: I would have used `response.should.be('string')` here, but response doesn't even have the `should` method, because it's a `String` :). Hence the `typeof response` workaround here.

Issue #349, opened 2016-04-25
(-20 // 7) gives -2, but the official specs says it should be -3


tjcegodk:
According to the [Twig specs](http://twig.sensiolabs.org/doc/templates.html#math):

>//: Divides two numbers and returns the floored integer result. {{ 20 // 7 }} is 2, {{ -20 // 7 }} is -3 (this is just syntactic sugar for the round filter).

But twig js gives -2, which is the result of truncating any number past the . (arguably also wrong in that floating point numbers could be 0.99999999 which should probably be treated as one).

I was able to fix it by replacing `parseInt` with `Math.floor` in `twig.expression.operator.js`

(no PR, sadly, because I need permission from my employer).

dave-irvine:
Thanks I'll try and write a fix for this based on your suggestion.

PR #353, opened 2016-05-03
Fix negative number parsing and // flooring

dave-irvine:
Turns out we completely fail to parse negative numbers!


dave-irvine:
Fixes #349

PR #378, opened 2016-05-31
Change the order of the expression.

chenos:
Although ignore missing not working

```
{% include 'demo.html' ignore missing with {'foo': 'bar'} only %}
```

not

```
{% include ignore missing 'demo.html' with {'foo': 'bar'} only %}

```

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,8 +1,8 @@
 demos
 tools
 test
-src
 qtest
+src
 .hg
 .git
 .gitignore
--- a/.npmignore
+++ b/.npmignore
@@ -1,6 +1,7 @@
 demos
 tools
 test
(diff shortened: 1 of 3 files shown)


dave-irvine:
We don't bundle src/ in npm, that is for the final package distribution, so if you could revert that, and your change to package.json I can take another look. Thanks!

chenos:
I have restored the package.json and .npmignore

dave-irvine:
Do you think you could add a test case that fails before your changes and works after your changes? This helps us make sure we don't break the thing you fixed in the future!

dave-irvine:
Was fixed in https://github.com/twigjs/twig.js/pull/424

Issue #374, opened 2016-05-27
Bower distribution missing


SirWaddles:
After updating from an older version to 0.9.5, through the bower package manager, there doesn't appear to be anything to load or require into JavaScript.

I'm not sure if I'm doing anything wrong, my bower json configuration is just

```
"twig.js": ">=0.5.8",
```

As far as I can tell, the bin, lib and src directories are ignored from bower.json, and this wasn't a problem when twig.js and twig.min.js were being built into the main directory, but that doesn't appear to be the case any more.

Is bower not supported any more? Or am I not configuring something correctly?

Thanks in advance.

dave-irvine:
So I haven't been checking whether the bower distribution works. I wasn't aware people still used bower really, I thought most people had switched to using npm for everything?

Seeing as I've just had two Issues opened regarding Bower I guess people do still use it!

I'll investigate how to bundle things for Bower, but its not something I've ever done, and I really don't fancy complicating our release process much more.

Blueblau:
I use bower along side npm, for things that also needs CSS bundled with it. So I would love if this would work as well 👍

dave-irvine:
twig.min.js is built when you run `npm run build-browser`.

tomfun:
Could anyone explain right way to use npm instead of bower please, sorry if this question is in wrong theme.

dave-irvine:
Ok the fact that I have to store the compiled and minified version in git pretty much seals this for me. If this was a compiled language and I had to store my binary in git for bower to work, that would never fly.

I'll let you do your own searches for this, but I don't seem to be the only one who thinks this way: https://www.google.co.uk/search?q=use+npm+instead+of+bower

If someone wants to start another repository that takes the current twig.js npm package, runs the build job and then publishes to bower, go ahead, but this isn't something I want to maintain.

@Blueblau To get css bundles as well, have you considered Webpack?

olets:
@SirWaddles just be clear: you aren't still just getting 0.5.8? Don't know if Bower's added extended support for semver matching, but in my understanding of npm updates to versions under 1.0.0 won't be picked up (that is, for example `^` won't work until v1.0.0)

SirWaddles:
[...]

PR #380, opened 2016-06-01
build for bower, update bower.json

tomfun:
when I download package via bower, I can't use it, because there is no compiled file for browser 'twig.min.js'.  
This file must be a part of git repository for bower.


tomfun:
https://github.com/twigjs/twig.js/issues/374

tomfun:
ok, but it is sad, there is no other way. Then I recomend remove bower.json file at all, if you don't support it any more.

PR #411, opened 2016-08-27
Updated readme with extended use case

tiborsaas:


r3wt:
Thanks, looks good to me

Issue #393, opened 2016-06-30
Bower installation error.


Rafaelgfirmino:
In bower installation dont exists /src/twig.js

olets:
#374 has the discussion about how/if bower should be supported

dave-irvine:
Removed bower.

PR #420, opened 2016-10-05
Remove Bower support

### LINKED ISSUE #417: Problem installing from bower
When run bower install twig.js not download any twig.min.js, only documentation and example files.

besologic:
Based on https://github.com/twigjs/twig.js/issues/374#issuecomment-222958349 it looks like there isn't an interest in maintaining Bower support from the maintainers. This PR removes `bower.json` from the repository and updates the README accordingly.

If at some point this changes, Bower support can be re-added. I personally just tried using Bower thinking it worked. Bower support looks like it's been busted since May 2016, this would alleviate any further confusion.

Resolves #417, #393, and #374


besologic:
After merging, https://github.com/twigjs/twig.js/wiki/ should be updated to remove Bower references as well.

Issue #405, opened 2016-08-02
Cannot clone repo on Windows


dave-irvine:
As reported in #402, it is impossible to clone the repository on Windows. This is because of a : in a filename.

PR #429, opened 2016-11-21
Adding support of 'same as'

oleg-andreyev:
`sameas` is deprecated
Ref.: https://github.com/twigphp/Twig/blob/1.x/doc/deprecated.rst

### DIFF
--- a/src/twig.tests.js
+++ b/src/twig.tests.js
@@ -39,7 +39,7 @@ module.exports = function (Twig) {
         },
         sameas: function(value, params) {
             console.warn('`sameas` is deprecated use `same as`');
-            return Twig.test['same as'](value, params);
+            return Twig.tests['same as'](value, params);
         },
         iterable: function(value) {
             return value && (Twig.lib.is("Array", value) || Twig.lib.is("Object", value));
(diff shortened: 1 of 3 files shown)


src/twig.tests.js
```diff
@@ -1,60 +1,64 @@
...
+        },
+        'null': function(value) {
+            return this.none(value); // Alias of none
+        },
+        'same as': function(value, params) {
+            return value === params[0];
+        },
+        sameas: function(value, params) {
+            console.warn('`sameas` is deprecated use `same as`');
+            return Twig.test['same as'](value, params);
```
dave-irvine: I think this should be:

`return Twig.tests['same as'](value, params);`

dave-irvine:
This breaks the namespaces test. https://github.com/twigjs/twig.js/pull/429/commits/ec475508128e1a185c661df4a42378c9ab4ec131

Can you remove that commit from this pull request and open a seperate pull for fixing #405?

oleg-andreyev:
I'm in process of fixing it. I'd leave that commit because it blocks
development of Windows machine

dave-irvine:
Sure I understand it blocks development, I would still like your fix, just in a seperate pull request. One pull request per issue fixed :)

oleg-andreyev:
@dave-irvine okay, will do

PR #430, opened 2016-11-22
Fix windows colon colon namespace

oleg-andreyev:
Fix for #405


dave-irvine:
You'll also need to fix https://github.com/twigjs/twig.js/blob/master/test/test.namespaces.js#L8 as it tries to load the old file :)

oleg-andreyev:
@dave-irvine should be fixed

oleg-andreyev:
@dave-irvine you're awesome, thanks for fast response 👍 :)

PR #438, opened 2016-12-19
Fix 'from' tag logic and corresponding test

yoozze:
Aliases in `from` tag are ignored.
Test is passing, because it isn't actually testing for aliases.

This PR fixes:
- regular expressions for parsing aliases in `from` tag
- `from` tag test template to actually test for aliases


dave-irvine:
Thanks!

Issue #426, opened 2016-11-02
Twig Sync Loading - In Browser?


cwbeck:
Extending a template works in Chrome, but does produce the following error:

<img width="693" alt="screen shot 2016-11-02 at 11 18 57" src="https://cloud.githubusercontent.com/assets/425262/19926807/9e1fa124-a0ee-11e6-8221-8c90d7c9397a.png">

This should be all async, so why is this blocking? Anyway to fix this?

```
var template = twig.twig({
    id: "login",
    namespaces: { 'views_dir': 'js/app/' },
    href: "js/app/pages/login/login.twig",
    async: true,
    load: function(template) {
        console.log(template.render());
    }
});
```

Where js/app/pages/login/login.twig -> extends another template

DrParanoia:
I can confirm this.

The first template is loaded asynchronously, however if you have any includes that requires additional templates to be loaded, these templates are loaded in a synchronous manner. To make matters worse it freezes the UI until templates are loaded.

dave-irvine:
Can you give me an example of the two templates? Just minimum breaking example.

cwbeck:
Any template extending another template. Where first template, in my case ```login.twig``` can just contain:

```
{% extends "base.twig" %}
...
```

And `base.twig` contains:

```
<b>hello world!</b>
```

It forces a sync request for subsequent templates rather than a fetch as an async waterfall which is the desired pattern and thus not blocking the page. templates must be fetched in order as dependencies are unknown until point of render on each.

Hope that helps.

cwbeck:
Is this still an active project?

dave-irvine:
As much as it can be with one maintainer, yes. Happy for a pull request.

DrParanoia:
When async loading the template, all included or extended templates that it has will not load in async mode.

I am afraid that fixing this is not a trivial task :/ But I just have to ask - is there any hope that this will be fixed in the near future?

dave-irvine:
I'd love to fix it, but I'm not currently using Twig in any of my projects at my job, or any of my spare-time projects. As such I don't have much time to devote to twig.js other than attending to pull requests and cutting releases.

If someone else can fix this I'm more than happy to help get it landed, but I don't have any time to dedicate to it at the moment.

I think @olets uses Twig.js for some of his work projects, he might be a candidate for fixing :)

cwbeck:
[...]

PR #439, opened 2016-12-26
Replace `ChildContext` with simple object

willrowe:
### Issue
While I was working on a PR which affects the contexts and merging them together, I kept running into issues with inconsistency in relation to how variables in the contexts were being stored. This resulted in it being difficult for two contexts to be merged together. It turned out to be an issue with how a new `ChildContext` stores passed variables on the prototype. Depending on how it is handled, the variables can become essentially hidden and not carry through when, for example, extending templates.

### Solution
This PR removes `ChildContext` completely, replacing it with a simple object which can be merged with other contexts using `Twig.lib.extend()`. The only thing that I can determine that `ChildContext` is used for, is to clone the original context so that any changes are isolated in the child. Every instance where `ChildContext` is used has been replaced with a call to `Twig.lib.copy()` which returns a new simple object with the same variables.

### Considerations
This would make working with the contexts much simpler, but I may be missing some details as to why `ChildContext` was created in the first place. I would greatly appreciate any background on it which would improve this PR or point out why `ChildContext` is needed.

### DIFF
--- a/src/twig.logic.js
+++ b/src/twig.logic.js
@@ -260,7 +260,7 @@ module.exports = function (Twig) {
                     },
                     // run once for each iteration of the loop
                     loop = function(key, value) {
-                        var inner_context = Twig.ChildContext(context);
+                        var inner_context = Twig.lib.copy(context);
 
                         inner_context[token.value_var] = value;
 
@@ -600,8 +600,7 @@ module.exports = function (Twig) {
                 return token;
             },
             parse: function (token, context, chain) {
-                var template,
-                    innerContext = Twig.ChildContext(context);
+                var template;
                 // Resolve filename
                 var file = Twig.expression.parse.apply(this, [token.stack, context]);
 
@@ -616,10 +615,7 @@ module.exports = function (Twig) {
                 }
 
                 // Render the template in case it puts anything in its context
-                template.render(innerContext);
-
-                // Extend the parent context with the extended context
-                Twig.lib.extend(context, innerContext);
+                template.render(context);
 
                 return {
                     chain: chain,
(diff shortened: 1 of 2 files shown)


dave-irvine:
`ChildContext` was written before my time unfortunately, so I don't really know what its purpose is as opposed to a plain object.

Unfortunately as it is fairly key to parsing, I think if we replace it we will need to bump the major version, even though all current tests pass.

I'd love to get #426 fixed before we cut a new major version so that browser support feels a bit more complete. Apparently twig.js has stopped working in jsfiddle which I assume means it has stopped working in browsers entirely.

src/twig.logic.js
```diff
@@ -616,10 +615,7 @@ module.exports = function (Twig) {
                 }
 
                 // Render the template in case it puts anything in its context
-                template.render(innerContext);
-
-                // Extend the parent context with the extended context
-                Twig.lib.extend(context, innerContext);
```
dave-irvine: So I had this extend in for a reason, see https://github.com/twigjs/twig.js/pull/369.

What did you manage to do instead to make the tests pass?

willrowe:
Thanks for taking a look. This is definitely an exploratory improvement, so waiting until the next major version makes sense.

src/twig.logic.js
willrowe: Since the context object is passed by reference (instead of a copy), any changes made to the context when rendered are made on the same instance. This made the call to `Twig.lib.extend()` extraneous, so I removed it. #369 added the `template.render()` call which applies any changes to the context and is retained below.

Issue #357, opened 2016-05-06
Feature request: Asynchronous function callback


rolyataylor2:
So what I am needing and I have been unable to figure out with the twig source is to allow for:

```
Twig.extendFunction("DatabaseCall", function (arguments) {
    return 'Template Stuff';
});
```

The problem is that node wants to do everything via callbacks... So what would be neat is to extend twig using a function that upon completion it does not continue rendering but instead waits for a callback... So like this...

```
Twig.extendFunctionAsync("Import", function (callback, name, twigVars) {
    // Pull database things
    Callback(ReturnVariable);
});
```

The exact problem I am having is I want to allow template creators to render items from the database using that function but to pull from the database it needs to be Async because node cannot do network calls synchronously.

I will keep plugging away at this but you guys have more insight as to how your renderer works... 

Thanks for reading 👍

dave-irvine:
So I think what you're trying to do is probably similar to: https://github.com/twigjs/twig.js/issues/61, is that right?

rolyataylor2:
I did not see that one. That is what I am intending to do because of the nature of a mysql call in node js.

dave-irvine:
Its definitely a cool idea, but unfortunately I really can't prioritize it at the moment because its not something you can do in Twig PHP. I'm currently aiming to try and get more compatibility with Twig PHP as my priority.

rolyataylor2:
Makes sense. I am thinking of maybe rendering the template and using a randomized token as a placeholder for the variable then after the render is done going though and replacing the tokens with the returned variables. So kinda like a double render or a basic .replace('DBVAR','info') kind of thing.

kachurun:
Why it not work with {% for %} tag?

Simple example:
Function:

```
Twig.extendFunction('async_test', (id) => {
    return Promise.resolve(`Test: ${ id }`);
});
```


Template: 
```
<div id="single>
    {{ async_test(1000) }}
</div>

// ids is [2000, 20001, 20002]
<div id="each">
    {% for id in ids  %}
        <div data-index="{{ loop.index }}">
            {{ async_test(id) }}
        </div>
    {% endfor %}
</div>
```

returns 
```
<div id="single>
    Test: 1000
</div>

<div id="each">
    
</div>
```

where #each is empty. 

How i can fix it?

willrowe:
Duplicate of #61.

PR #457, opened 2017-03-01
Asynchronous rendering and function/filter promises

JorgenEvens:
Allows for functions and filters to return promises and render asynchronously (https://github.com/twigjs/twig.js/issues/61).

I've tried to build this PR based on the discussions in the original issue and some offline discussions with @Gertt. The currently implementation tries to change as little of the existing code as possible, but I had to restructure the implementation of some `logic`/`expression` code such that it would behave correctly when used with promises.

Only the `Twig.expression.parse`, `Twig.expression.resolve`, `Twig.parse` and `Twig.Template.render` methods are truly aware of the fact that promises are used internally. If the flag `allow_async` is set these methods will allow functions and filters to behave asynchronously, otherwise it will throw as soon as the asynchronous behaviour is detected.

The original synchronous behaviour of Twig.js is preserved using a internal Promise implementation (`Twig.Promise`) that does not use `setTimeout` to call the `then` methods, in effect running all code synchronously. Since we can mix different implementation of promises (any thenable really), we can return regular promises from functions or filters at which time rendering will start to run asynchronously.

The external API should be left unchanged as far as I can tell. To allow for asynchronous rendering a user would use `template.renderAsync` which will return a promise.

Internally Twig.js should use the `Twig.expression.parseAsync`, `Twig.expression.resolveAsync`, `Twig.parseAsync` and `template.renderAsync` to ensure that every component is ready to receive a promise. These methods are shorthands for an easier transition that take the exact same arguments as the regular methods, but will automatically set the `allow_async` argument to true.

What are your thoughts on this? I'm happy to give this a few revisions if you'd like to try out some different things.

Some things that might improve readability of this PR:

- Viewing this PR without indendations highlight (adding `?w=1` to the URL). https://github.com/twigjs/twig.js/pull/457/files?w=1
- [ES6 arrow functions](http://babeljs.io/docs/plugins/transform-es2015-arrow-functions/), less `var that = this` and generally more concise methods inside of `.then(...)`. Depending on your target this would require `babel`.

### DIFF
--- a/ASYNC.md
+++ b/ASYNC.md
@@ -1,0 +1,218 @@
+# Twig Asynchronous Rendering
+
+## Synchronous promises
+
+The asynchronous behaviour of Twig.js relies on promises, in order to support both the synchronous and asynchronous behaviour there is an internal promise implementation that runs fully synchronous.
+
+The internal implementation of promises does not use `setTimeout` to run through the promise chain, but instead synchronously runs through the promise chain.
+
+The different promise implementations can be mixed, synchronous behaviour however is no longer guaranteed as soon as the regular promise implementation is run.
+
+### Examples
+
+**Internal (synchronous) implementation**
+
+[Internal implementation](https://github.com/JorgenEvens/twig.js/tree/master/src/twig.async.js#L40)
+
+```javascript
+console.log('start');
+Twig.Promise.resolve('1')
+.then(function(v) {
+    console.log(v);
+    return '2';
+})
+.then(function(v) {
+    console.log(v);
+});
+console.log('stop');
+
+/**
+ * Prints to the console:
+ * start
+ * 1
+ * 2
+ * stop
+ */
+```
+
+**Regular / native promises**
+
+Implementations such as the native promises or [bluebird](http://bluebirdjs.com/docs/getting-started.html) promises.
+
+```javascript
+console.log('start');
+Promise.resolve('1')
+.then(function(v) {
+    console.log(v);
+    return '2';
+})
+.then(function(v) {
+    console.log(v);
+});
+console.log('stop');
+
+/**
+ * Prints to the console:
+ * start
+ * stop
+ * 1
+ * 2
+ */
+```
+
+**Mixing promises**
+
+```javascript
+console.log('start');
+Twig.Promise.resolve('1')
+.then(function(v) {
+    console.log(v);
+    return Promise.resolve('2');
+})
+.then(function(v) {
+    console.log(v);
+});
+console.log('stop');
+
+/**
+ * Prints to the console:
+ * start
+ * 1
+ * stop
+ * 2
+ */
+```
+
+
+## Async helpers
+
+To preserve the correct order of execution there is an implemenation of `Twig.forEach()` that waits any promises returned from the callback before executing the next iteration of the loop. If no promise is returned the next iteration is invoked immediately.
+
+```javascript
+var arr = new Array(5);
+
+Twig.async.forEach(arr, function(value, index) {
+    console.log(index);
+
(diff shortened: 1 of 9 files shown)


dave-irvine:
Well that is some fantastic work, well done! It'll take me a little while to go over it.

Can you just step me through how https://github.com/twigjs/twig.js/pull/457/files#diff-22566ec93dc04f3aa4f038de2fd55f29R771 is behaving if you want synchronous behaviour? Or rather, how it is switching between the two?

Additionally, how do you configure Twig to be either sync or async?

JorgenEvens:
`Twig.async.forEach` will loop over each item in the array, if nothing is returned from the callback, the callback will be immediately invoked for the next iteration. If however the callback returns a promise it will wait until that promise resolves before invoking the next iteration.

Because the internal Promise implementation does not use a `setTimeout` to call the next method in the promise chain the entire promise chain should resolve before reaching any [checks](https://github.com/twigjs/twig.js/pull/457/files#diff-22566ec93dc04f3aa4f038de2fd55f29R836) on `is_async` which is set to `false` in the [last](https://github.com/twigjs/twig.js/pull/457/files#diff-22566ec93dc04f3aa4f038de2fd55f29R815) `.then(..)` call.

As soon as one of the promises returned from inside of Twig (function / filter) is a regular Promise (not our internal implementation) that promise's internal implementation [will use](https://www.promisejs.org/implementing/) a `setTImeout` (I believe this is required by spec) and `is_async` will not be set to `false` before we reach the check on `is_async`.

Switching between the two behaviours is accomplished using the `allow_async` argument that is passed to the different `.parse()` methods. If a template would use an asynchronous function while `allow_async` is not set it will throw an error to notify the user of this.

The initial location where `allow_async` is set to true is in the `template.renderAsync` method, or any time any of the `.parseAsync()` shorthands are called.

Something I think that might need some clarification after writing this is that `renderAsync` and the `parseAsync` methods might return and resolve a promise **synchronously** due to our internal Promise implementation, but it will **always** return a promise. Only when a truly asynchronous function or filter is used will the promise returned by Twig resolve asynchronously.

dave-irvine:
Maybe a wiki page to go with this or a README.md entry on how async can be used?

JorgenEvens:
Sure, I'll see what I can do this evening.

JorgenEvens:
I've added some documentation on the implementation and usage, is [this](https://github.com/JorgenEvens/twig.js/blob/5cb8f79a7a3b3c5ded8f8f7e605891577c34dc55/ASYNC.md) what you had in mind?

dave-irvine:
Wow. Just wow. By far and away the best pull request I have ever had the pleasure of merging.

Thank you for your hard work.

JorgenEvens:
No problem, it was my pleasure.

Do you have any thoughts on adding `babel` to the build step to clean up the `var that = this`, and make this a little easier to maintain in the future?

I think it might also be wise to add linting (my preference would be `eslint`).

What do you think? I might create a separate PR for those if you'd like that.

dave-irvine:
I would like to start implementing as ES6 so babel would be a good step, also happy to take on eslint using the AirBNB defaults but I imagine there are going to be *lots* of errors.

As you can see from the Issues list, there's a lot of things that need fixing up. Maybe eslint will catch something that accidentally fixes something!

Gertt:
@dave-irvine I think this PR also resolves a few outstanding issues:

https://github.com/twigjs/twig.js/issues/357
https://github.com/twigjs/twig.js/issues/61

And maybe we should let the guys at nodefony (https://github.com/nodefony/nodefony/issues/35) know about this, since they seem to be active on this in the last week?

Issue #462, opened 2017-03-19
Template id not returned as part of the error when an exception is thrown


ericmorand:
Both Twig.compile and Twig.parse fail at returning the id of the template where the error occurred when an exception is thrown. For example, calling an unknown filter 'bar' will throw the following exception:

`Unable to find filter bar`

Pretty useless without the name of the template where the filter is called.

This makes things even worse with middlewares that watch files based on dependencies instead of globs: those templates are never tracked by the middleware and thus fixing the error never leads to a new build round trip.

PR #464, opened 2017-03-19
Fix issue #462 - Template id not returned as part of the error when an exception is thrown

ericmorand:
This commit fixes the issue by adding a 'file' property to Twig.Error and making sure that it is always set to the path of the file where the error is located.

Test suite has been updated accordingly.


test/templates/error/compile/entry.twig
```diff
@@ -0,0 +1,2 @@
+Should throw a compile error
+{{ "foo" bar }}
```
dave-irvine: Can you add the newlines at the end of these files?

test/test.rethrow.js
```diff
@@ -1,23 +1,78 @@
...
+
+delete require.cache[path.resolve(path.join(__dirname, '../twig.js'))];
+
 var Twig = Twig || require("../twig"),
     twig = twig || Twig.twig;
 
-describe("Twig.js Rethrow ->", function() {
-    it("should throw a \"Unable to parse 'missing'\" exception", function() {
-        (function(){
+describe("Twig.js Rethrow ->", function () {
```
dave-irvine: I think the rest of our tests use `function()` rather than `function ()`. Can you update accordingly?

ericmorand: Oh yes, sorry about this. I fix this right now.

test/templates/error/compile/entry.twig
ericmorand: Yep.

ericmorand:
@dave-irvine That should be OK. Let me know if you want some more changes.

Issue #461, opened 2017-03-17
cordova app prodblem android4.4 chrome 30 WebView


stripcode:
#####webpack2
var path = require('path');

module.exports = {
  entry: './static/app.js',
  output: {
    filename: 'app.js',
    path: path.resolve(__dirname, 'www/js')
  },
  devtool: "source-map",

  module: {
    rules: [{
      test: /\.js$/,
      exclude: /node_modules/,
      loader: "babel-loader",
      query: {
        presets: ["env"],
      }
    }]
  },

  resolve: {
    modules: [path.resolve(__dirname, "static"), "node_modules"]
  }
};

##### app.js

import Twig from "twig"

Renderer.render = function(templateUrl, data){
  var template = Twig.twig({
    href: templateUrl,
    async: false
  });
  return template.render(data);
};

on android4.4 WebView error in console
Uncaught SyntaxError: Use of const in strict mode. 

	    Twig.Promise.all = function(promises) {
	        const results = [];

	        return Twig.async.forEach(promises, function(p, index) {
	            if (!Twig.isPromise(p)) {
	                results[index] = p;
	                return;
	            }

	            return p.then(function(v) {
	                results[index] = v;
	            });
	        })
	        .then(function() {
	            return results;
	        });
	    };

I change const to var and have success.
Why module with const and not transpile

dave-irvine:
Fixed in #470

Issue #470, opened 2017-04-04
Unexpected keyword 'const'. Const declarations are not supported in strict mode.


mrksmts:
This is for Safari 9.1.2 and relates to #461 

```
SyntaxError: Unexpected keyword 'const'. Const declarations are not supported in strict mode.
(anonymous function) — twig.js:7453
```

``` javascript
Twig.Promise.all = function(promises) {
    const results = [];

    return Twig.async.forEach(promises, function(p, index) {
        if (!Twig.isPromise(p)) {
            results[index] = p;
            return;
        }

        return p.then(function(v) {
            results[index] = v;
        });
    })
    .then(function() {
        return results;
    });
};
```

darrenmothersele:
Will there be a release with this fix? 
Thanks!

dave-irvine:
Done

PR #471, opened 2017-04-04
Fix issues #461 #470 - Unexpected keyword 'const'

mrksmts:
Replaced const with var


mrksmts:
With the speed of light, thanks @dave-irvine

Issue #472, opened 2017-04-07
Async template loading killed the performance...


DrParanoia:
Hey there! Thanks for adding async template loading in include tags.

The thing is, this feature killed the performance. (751a8ba41b4275ee480d9c6de87d221e052a4b17)
My templates (that are precompiled and don't use the async loading) that took ~150ms to render now takes ~500ms. That's almost 3x times the difference.

I guess it's because the render loop is not optimised now (because of the promises).

I found out that it is much better to precompile the templates and just include them like simple JS scripts.

In ideal world the async loops should be used only if there is something to include and if the included template is not already cached. Any thoughts?

DrParanoia:
Is anyone here? :/

willrowe:
I would like to see some benchmarking setup for this so different changes can be compared.
Any ideas for tools that can be used for that?

DrParanoia:
Unfortunately I don't know of any tools like that, but you can just have two tabs open with different twig.js files imported. Or you're looking for a template example that is known to work slow with the new version?

JorgenEvens:
Hi @DrParanoia,

We have also experienced some slow rendering templates, so let me check what is going on exactly.

Only using the async loops when there is async functionality included would be the ideal situation, but we can't reliably detect this in advance without break the existing API.

The promises used by this implementation don't follow the spec in that they don't call the next function on the next tick. So while it would add some overhead, it should not add the amount you ( and we ) are seeing.

JorgenEvens:
Did some digging and I was not experiencing slow rendering templates, but had some other code that ran right before it being slow.

That being said, if you could share a template on which you are experiencing slow rendering I'd be happy to have a look.

DrParanoia:
Hi there! Sorry for the wait. I think your test case is too simple to see the issue. I've just tested your commit but the issue is still there. I've used this code just for testing purposes.

```
var template = "{{ echo_test }}\
{% for item in items %}\
    <h1>{{ item.title }}</h1>\
    {% if item.is_active %}\
        <h3 style=\"color: green;\">Active</h3>\
    {% else %}\
        <h3 style=\"color: red;\">Inactive</h3>\
    {% endif %}\
{% endfor %}";

console.time('twig');
for(var i = 0; i < 1000; i++) {
    Twig.twig({
        data: template
    }).render({ 
        echo_test: 'Data for echo',
        items: [
            {
                title: "This is a title",
                is_active: true
            },
            {
                title: "This is a title",
                is_active: false
[...]

PR #486, opened 2017-06-16
Rendering performance improvements

JorgenEvens:
As a result of https://github.com/twigjs/twig.js/issues/472 I've been looking into improving rendering performance. These changes reduce rendering times back to pre-async performance numbers.

**Updates**

- [Promise] Reduced the amount of closures used by a promise.
- [Promise] Reduced branching by replacing functions based on promise state.
- [Promise] Use numeric constants to check promise state.
- [Promise] Lazily convert `handlers` to an array.
- [Promise] Use `Twig.attempt` instead of `try/catch` ([optimization-killer](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#2-unsupported-syntax))
- [Promise] Move `append` out of constructor, js engine only has to optimize it once.
- [async/forEach] Reduce amount of closures.
- [async/forEach] Create less objects in memory.
- [Twig.compile] Use `Twig.attempt` instead of `try/catch`
- [Twig.expression.tokenize]: Don't leak `arguments` ([optimization-killer](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#32-leaking-arguments))

**Average rendering durations on different branches over 100 runs:**

These numbers are using some of our own internal templates, feel free to check this branch against your own.

- Current master: 0.422ms
- v0.10.3 ( pre-async ): 0.356ms
- This PR: 0.321ms

**Example of test script**

```js
#!/usr/bin/env node

const Twig = require('../twig');

let i = 100;
const duration = [];

while( i-- ) {
    console.time('twig');

    Twig.twig({
        data: 'Hello {{world}}'
    }).render({ world: 'World' });

    console.timeEnd('twig');
}
```

### DIFF
--- a/src/twig.async.js
+++ b/src/twig.async.js
@@ -54,6 +54,38 @@ module.exports = function (Twig) {
         return handlers;
     }
 
+    Twig.Thenable = function(then, value, state) {
+        this.then = then;
+        this._value = state ? value : null;
+        this._state = state || STATE_UNKNOWN;
+    }
+
+    Twig.Thenable.prototype.catch = function cc(onRejected) {
+        return this.then(null, onRejected);
+    }
+
+    Twig.Thenable.resolvedThen = function(onResolved) {
+        // Shortcut for resolved twig promises
+        if (typeof onResolved != 'function')
+            return Twig.Promise.resolve(this._value);
+
+        var self = this;
+        return Twig.attempt(function() {
+            return Twig.Promise.resolve(onResolved(self._value));
+        }, Twig.Promise.reject);
+    }
+
+    Twig.Thenable.rejectedThen = function(onResolved, onRejected) {
+        // Shortcut for rejected twig promises
+        if (typeof onRejected != 'function')
+            return Twig.Promise.reject(this._value);
+
+        var self = this;
+        return Twig.attempt(function() {
+            return Twig.Promise.resolve(onRejected(self._value));
+        }, Twig.Promise.reject);
+    }
+
     /**
      * An alternate implementation of a Promise that does not fully follow
      * the spec, but instead works fully synchronous while still being
@@ -62,7 +94,7 @@ module.exports = function (Twig) {
      * These promises can be mixed with regular promises at which point
      * the synchronous behaviour is lost.
      */
-    Twig.Promise = function(executor, /* internal -> */ initialState, initialValue) {
+    Twig.Promise = function(executor) {
         // State
         var handlers = null;
 
@@ -79,9 +111,6 @@ module.exports = function (Twig) {
         var append = function unknown(onResolved, onRejected) {
             handlers = pending(handlers, onResolved, onRejected);
         };
-
-        if (initialState)
-            append = initialState == STATE_RESOLVED ? resolved : rejected;
 
         function onReject(e) { changeState(STATE_REJECTED, e); }
         function changeState(newState, v) {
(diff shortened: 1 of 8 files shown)


DrParanoia:
Hi there! Sorry for the wait. I think your test case is too simple to see the issue. I've just tested your commit but the issue is still there. I've used this code just for testing purposes.

`var template = "{{ echo_test }}\
{% for item in items %}\
	<h1>{{ item.title }}</h1>\
	{% if item.is_active %}\
		<h3 style=\"color: green;\">Active</h3>\
	{% else %}\
		<h3 style=\"color: red;\">Inactive</h3>\
	{% endif %}\
{% endfor %}";

console.time('twig');
for(var i = 0; i < 1000; i++) {
	Twig.twig({
        data: template
	}).render({ 
		echo_test: 'Data for echo',
		items: [
			{
				title: "This is a title",
				is_active: true
			},
			{
				title: "This is a title",
				is_active: false
			}
		] 
	});
}
console.timeEnd('twig');
`

JorgenEvens:
The test case was simply an example of the test harness I used, I dropped in some of our internal templates.

That being said, I will try and see what I can accomplish using your template. There is definitely more room for improvement here.

JorgenEvens:
@DrParanoia Only a ~20ms difference after the last changes, will see if I can get it all the way back to original speeds.

JorgenEvens:
@DrParanoia On my machine the two implementation are now running at equal speed. Could you try this PR and see if it works ok for you?

I've used your testscript to generate these results, so you should be able to reproduce these.

# i=10; while [ $i -gt 0 ]; do test-05-v0.10.3.js; i=$((i-1)); done
twig: 261.706ms
twig: 250.302ms
twig: 246.580ms
twig: 249.145ms
twig: 246.331ms
twig: 245.665ms
twig: 258.785ms
twig: 256.907ms
twig: 253.217ms
twig: 247.564ms

# i=10; while [ $i -gt 0 ]; do test-05.js; i=$((i-1)); done
twig: 250.118ms
twig: 244.766ms
twig: 247.833ms
twig: 253.946ms
twig: 240.143ms
twig: 247.774ms
twig: 246.613ms
twig: 241.833ms
twig: 251.292ms
twig: 249.193ms

DrParanoia:
@JorgenEvens Great job! I've run my test template with 10k iterations 10 times with each twig.js version. And results are:

old = 10799ms
new = 11357ms (only 5.16% slower than the old one)

I can definetly use it with only 5% slowdown.

DrParanoia:
@JorgenEvens Wow! After your last commit, the new version is now about 10.72% faster than the old one! That's a really great job there! Thanks!

JorgenEvens:
@DrParanoia That is actually surprising as the last commit only improved performance marginally on my machine.

At this point I would still like to fix the usage of `apply` with `call` where the number of arguments are known in advance, apart from that I think I'm finished for now.

DrParanoia:
@JorgenEvens Thanks for your job. And as much as I hate to bring the bad news, there are still significant performance problems when using `include` even if the templates are already compiled.

I've made this simple copy/paste into Chrome console test:
```
window.TwigTemplates = window.TwigTemplates || {};
if(!window.TwigTemplates["views/project/test_template_1.twig"]) {
    window.TwigTemplates["views/project/test_template_1.twig"] = Twig.twig({ autoescape: true, id: "/templates/get/?name=project/test_template_1.twig", namespaces: {"project":"/templates/get/?name=project"}, data: [{"type":"logic","token":{"type":"Twig.logic.type.include","only":false,"ignoreMissing":false,"stack":[{"type":"Twig.expression.type.string","value":"@project/test_template_2.twig"}]}}] });
    window.TwigTemplates["views/project/test_template_2.twig"] = Twig.twig({ autoescape: true, id: "/templates/get/?name=project/test_template_2.twig", namespaces: {"project":"/templates/get/?name=project"}, data: [] });
}

console.time('twig');
for(var i = 0; i < 100000; i++) {
    window.TwigTemplates["views/project/test_template_1.twig"].render();
}
console.timeEnd('twig');
```

Basically it's just two complied empty templates:
test_template_1.twig
```
{% include "@project/test_template_2.twig" %}
```
And an empty template test_template_2.twig. I've run this test 10 times for each twig version and the performance difference is really dramatic:
- old twig: 5292ms total
- new twig: 9112ms total (~72.18% slower)

JorgenEvens:
@DrParanoia I'll give that a try later tonight.

I did see that loading templates was far from optimal, but did not link it to this issue since it's behaviour should not have changed much by the inclusion of the async API.

DrParanoia:
@JorgenEvens Maybe performance drops because as soon as there is an include, the rendering switches to full async loop instead of tight one-threaded loop? Sorry if I don't make sense, haven't dug into your code :(

JorgenEvens:
Non of the twig.js internals cause it to switch to full async mode. And since you are calling `render` and not `renderAsync` that would throw an error anyway.

DrParanoia:
@JorgenEvens Hmmm... then I'm currently out of ideas :( But the templates are empty, so it's definitely something with the include mechanism. I have quite a few includes in my project, so I guess that's why I saw the decrease in performance in the first place.

JorgenEvens:
@DrParanoia just checked that I can reproduce your issue, so I will be looking into this tonight.

DrParanoia:
@JorgenEvens Thanks!

JorgenEvens:
@DrParanoia Just an update for you.

I have not yet nailed down the reason why, but these methods are many times slower now that they live inside of a `then` callback.

https://github.com/twigjs/twig.js/blob/master/src/twig.logic.js#L777-L794
Moving away from `try/catch` towards `Twig.attempt` only makes the code run slower which is weird.

https://github.com/twigjs/twig.js/blob/master/src/twig.logic.js#L796-L799
Really weird one, but takes up a ton of CPU time, despite it only creating a new object.

https://github.com/twigjs/twig.js/blob/master/src/twig.path.js#L21-L41
This method was slow before as well, but it seems slower than it actually should be.

DrParanoia:
@JorgenEvens Hey there, thanks for the update. Could this be related to that inside of `Twig.logic.type.include` no matter what `Twig.expression.parseAsync` and `template.renderAsync` are always called?

JorgenEvens:
`parse` used to be called as well as can be seen on this screenshot taken from the old version executing, and `parseAsync` is essentially only a passthrough that sets one of the arguments to a fixed value.

```
Twig.logic.parseAsync = function (token, context, chain) {
        return Twig.logic.parse.call(this, token, context, chain, true);
    }
```

I personally think there is some weird de-optimization / suboptimal optimization going on in the function due to the variables that are used outside of the closure. Especially the object allocation taking 30ms on a 900ms run is making me suspicious.

screenshot 2017-06-28 01 21 53

JorgenEvens:
This is what that section looks like using the new version.

screenshot 2017-06-28 10 57 38

DrParanoia:
Wow, now that's some overhead...

PR #524, opened 2018-01-05
Add allowable tags parameter to strip tags filter

isvaljek:


olets:
Currently the `striptags` filter ignores the allowed tags parameter. Per Twig docs,

```
{{ some_html|striptags('<br><p>') }}
```

should preserve all `<br/>`, `<br>`, `<p>`, `</p>`.

I tested this fix and it does work.

@isvaljek can you add tests for this?

isvaljek:
I wrote a couple of tests, but I'm still feeling fluefied, please check.

isvaljek:
Is there somthing else that needs to be done here?

PR #526, opened 2018-01-23
Update source() to not use __dirname

RobLoach:
__dirname represents the path to the current Node.js module (twigjs/src/twig.functions.js), not the currently running Twig script. See the documentation on __dirname for more info:
https://nodejs.org/docs/latest/api/modules.html#modules_dirname


dave-irvine:
@RobLoach tests please and I'll merge!

RobLoach:
I think the tests are already there. It's inherent across all the existing ones.

dave-irvine:
@RobLoach if there were already tests, surely they would be failing in the cases that your fix fixes?

RobLoach:
@dave-irvine Got it! Needed to do a bit more debugging to accurately demonstrate the issue.

Issue #442, opened 2017-01-10
Namespaces with source() do not resolve


ryuran:
Hello

When I include a partial with include tag it's work really well

```
  {% include '../partials/_exemple.twig' %}
```

But when I use source tag with same path, 
```
  {{ source('../partials/_exemple.twig') }}
```
It displays :
> Template "../partials/_exemple.twig" is not defined.

It should display partial content without interpreting it.

mattkenefick:
This is because of 4012 in twig.js. It's asking for:

    if (isNodeEnvironment) {
        loader = 'fs';
        path = __dirname + '/' + name;
    } else {
        loader = 'ajax';
        path = name;
    }

Which then gives you: `/Users/xyz/your_application/bin/node_modules/twig` 

It should be using the path shared by Node's Twig implementation.

chrisvanpatten:
I'm encountering the same issue.

idflood:
Here is the most simple implementation of the `source` function which worked for my use case. Eventually the `dir` will probably need to be adapted.

```
const dir = __dirname;
Twig.extendFunction('source', function(src) {
  return fs.readFileSync(dir + src, 'utf8');
});
```

RobLoach:
This may be fixed via https://github.com/twigjs/twig.js/pull/526 . Mind re-testing?

HakS:
Now source function works if you use a path relative to the root of the project, however the source function does not support namespaces as it does with Twig

Not sure if this issue should be open for this or if other issue should be created

renestalder:
Real bummer here. Does it take much effort to implement that? I can also create a PR if you point me to the gap in the code that has to be filled.

arismag:
bump, any update on this one?

PR #532, opened 2018-02-19
Add tests for namespaces in source function

charlestati:
When using the source function, namespaces are not working:

`{{ source("test::namespaces.twig") }}`

`Template "test::namespaces.twig" is not defined.`

Maybe I am doing something wrong? Are namespaces supposed to work with `includes` only?

### DIFF
--- a/test/templates/namespaces_@_source.twig
+++ b/test/templates/namespaces_@_source.twig
@@ -1,0 +1,1 @@
+{{ source("@test/namespaces.twig") }}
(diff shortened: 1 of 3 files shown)


RobLoach:
Don't think it's implemented...
https://github.com/twigjs/twig.js/blob/master/src/twig.functions.js#L245

willrowe:
This feature is not supported at this time. Please open an issue if you would like to see it added.

willrowe:
Actually there is already an issue for this: #442

PR #534, opened 2018-02-23
Fix bugs in the `embed` tag options

willrowe:
There were some issues with using the `ignore missing` option in conjunction with the `only` option.
It turned out to be due to the regex not matching the options correctly and leaving the token with some missing variables. The regex was a little greedy trying to match after each option against the end of the string, which is not necessary.

- This also adds a set of more robust tests for all the available options.

### DIFF
--- a/src/twig.logic.js
+++ b/src/twig.logic.js
@@ -1026,7 +1026,7 @@ module.exports = function (Twig) {
              *  Format: {% embed "template.twig" [with {some: 'values'} only] %}
              */
             type: Twig.logic.type.embed,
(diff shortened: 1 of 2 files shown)


willrowe:
Just a heads up, I'm also working on fixing another `embed` bug when used in a `for` loop.
I'm also working on re-implementing how blocks are handled throughout the entire library to fix some inconsistencies between this and the PHP version.

If these initial fixes I'm working on can be merged fairly quickly that will help immensely with re-implementing blocks.

Issue #352, opened 2016-05-03
Having some issues with embed tag


lvl99:
I'm doing a bunch of extending, including and embedding, and I'm finding when I use `embed` tags that it seems to ignore the content within them and further affects other block content within the same layout.

My structure is as follows:

```
  **layout.twig:**
    block header
      include partials/header.twig

    block content

    block footer
      include partials/footer.twig
```

I have another layout template which extends `layout.twig` and adds some additional blocks, as well as overwriting the `layout.twig`'s main header, content and footer blocks. In this layout, I'm using embeds in order to change specific areas within the header and footer:

```
  **dashboard.layout.twig extends layout.twig:**
    block header
      embed partials/header.twig
         (*contains block overrides for specific blocks within header.twig*)

    block content
       block dashboardContent

    block footer
      embed partials/footer.twig
        (*contains block overrides for specific blocks within footer.twig*)
```

Since the above is a reusable layout file, it's not one that I use for a specific page to be rendered to HTML. Here is the actual page's structure for rendering:

```
  **page.dashboard.layout.twig extends dashboard.layout.twig:**
    block dashboardContent
      (*custom content for this specific page*)
```

When the `page.dashboard.layout.twig` file is rendered, the header and the footer are unmodified (essentially same as default `layout.twig`) ignoring the embed block overrides, although it has rendered the `dashboard.twig` layout correctly, albeit with none of the content defined in the `dashboardContent` block.

When I remove the header/footer blocks or the `embed` tags from the `dashboard.layout.twig` structure, the `page.dashboard.layout.twig` block `dashboardContent` is correctly rendered for the page. Is it because I'm using two levels of extends that this might be causing some kind of issue? Are extends only designed for a single level of inheritance?

dave-irvine:
If it works in PHP twig, it should be working here. Does your example work with PHP twig? If so, could you try and simplify down to the simplest breaking case and reply with an example and we will try and figure it out.

lvl99:
I've made a quick version of the example I've given above. It seems to render correctly with all modifications cascading correctly: http://twigfiddle.com/vk11o4

I'm just going to run this example through Twig.JS now. Will report back soon.

lvl99:
It looks like Twig.JS doesn't seem to cascade correctly with the embeds:

![screen shot 2016-05-03 at 12 57 46 pm](https://cloud.githubusercontent.com/assets/238480/14981338/c70661e0-112e-11e6-9c60-2ba5f6d15d00.png)

[...]

PR #537, opened 2018-03-13
Fix embed blocks logic

toptalo:
I think that I fix issue #352


boschr:
@toptalo why is this PR closed? This solves also my issue with nested embeds. Like:

```
{% embed '_visual.twig' %}
  {% block visual_left %}
    {% embed '_content-block.twig' with { title: 1 } only %}
    {% endembed %}
  {% endblock %}

  {% block visual_right %}
    {% embed '_content-block.twig' with { title: 2 } only %}
    {% endembed %}
  {% endblock %}
{% endembed %}```

**Comment by toptalo** [comment]
@boschr because @willrowe working on complete re-implementation of how blocks are handled

https://github.com/twigjs/twig.js/issues/287#issuecomment-376200875

**Comment by boschr** [comment]
Thnx for the quick reply. I saw the comment of @willrowe. But in my opinion we can still do some parallel fixes for the current solution.

**Comment by toptalo** [comment]
I can reopen this PR, and maintainers will be decide what to do with it

**Comment by boschr** [comment]
Sounds like a plan. Probably they wil merge it sooner when you add an test. Really hope that we can merge fixes of this size fast (same for my own PR #556).

**Comment by toptalo** [comment]
@dave-irvine Hello! Can you help me to understand one moment?

Here is example from `test.embed.js` https://twigfiddle.com/1e50jc And it fails in twigfiddle.

I think code in `embed-include.twig` is wrong and embed blocks should not overwrite template blocks.

Am i right?

Also I found that empty block in `extender` template must clear `extendee` block
here is example: https://twigfiddle.com/vpprve
To inherit extendee block content we need to insert `{{ parent() }}` tag

**Comment by toptalo** [comment]
@willrowe is it ok if I copy-paste this two tests?
* `should support complex nested embeds`
* `should support multiple inheritance and embeds`

**Comment by willrowe** [comment]
Sure, use whatever you need.

File: test/test.embed.js [hunk 1]
```diff
@@ -105,4 +105,40 @@ describe("Twig.js Embed ->", function() {
...
+            data: '<{% block section_title %}section-title{% endblock %}><{% block section_content %}section-content{% endblock %}>',
+            id: 'section.twig'
+        });
+
+        twig({
+            allowInlineIncludes: true,
+            data: '{% extends "layout.twig" %}{% block body_header %}override-body-header{% endblock %}{% block body_content %}{% embed "section.twig" %}{% block section_content %}override-section-content{% endblock %}{% endembed %}{% endblock %}'
+        }).render().should.equal('<layout-header><override-body-header><section-title><override-section-content><layout-body-footer><base-footer>');
+    });
+});
```
willrowe: Please make sure there is a newline at the end of this file.

src/twig.logic.js
```diff
@@ -1148,17 +1148,21 @@ module.exports = function (Twig) {
                         }
                     }
 
+                    // store previous blocks
+                    that._blocks = Object.assign({}, that.blocks);
```
willrowe: Please use `Twig.lib.copy` to be consistent with the rest of the code.

src/twig.logic.js
```diff
@@ -1148,17 +1148,21 @@ module.exports = function (Twig) {
...
                     return Twig.parseAsync.call(that, token.output, innerContext)
                     .then(function() {
                         // render tempalte with blocks defined in embed block
-                        return template.renderAsync(innerContext, {'blocks':that.blocks});
+                        return template.renderAsync(innerContext, {'blocks': that.blocks});
                     });
                 })
                 .then(function(output) {
+                    // restore previous blocks
+                    that.blocks = Object.assign({}, that._blocks);
```
willrowe: Please use `Twig.lib.copy` to be consistent with the rest of the code.

willrowe:
This is good to go as a bugfix, if you could merge this @dave-irvine and tag v1.11.1 that would be great.

toptalo:
@willrowe thank you!

PR #538, opened 2018-03-14
[for] fixes empty loop output when containing async call

JorgenEvens:
For loops returned empty strings when using async rendering and async calls inside of the `for`-loop body.


dave-irvine:
8 minutes ago :)

PR #542, opened 2018-04-22
Upgrade to Webpack 4

mortenson:
I just started using Twig.js and wanted to make some updates to `webpack.config.js` when I realized that we're still on Webpack 1! We should upgrade to the latest version, 4, to open us up to new plugins and performance enhancements.

This doesn't improve anything about the build size right now, but the build time has been reduced about a second.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,3 @@
 language: node_js
 node_js:
-  - 4
   - 6
-
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ A loader is available from [zimmo.be](https://github.com/zimmo-be/twig-loader).
 
 ## Node Usage (npm)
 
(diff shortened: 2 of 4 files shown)


mortenson:
Tests are failing for Node 4 - we would have to change the minimum supported version of Node to 6.x for this PR to be merged.

Good news is that Node 4 is dropping out of LTS support at the end of this month (https://github.com/nodejs/Release#release-schedule), so there shouldn't be a reason to support it after that date.

dave-irvine:
@mortenson Ping me once 4 is EOL and I'll merge this.

mattgrill:
4.x end of life is today 🎉

mortenson:
@dave-irvine Ping 😉And thanks in advance!

Issue #287, opened 2015-12-22
Some context is lost with a combination of extends, embed, include


tremby:
I am losing some data variables in a certain situation on my site.

I gradually boiled the case down and have posted it in a gist: https://gist.github.com/tremby/62df667cf9fb9dc7a303

Pasting the output of the `test2page.twig` template here:

```
test2layout.twig 1      [foo]   [bar]
test2page.twig 1        [foo]   []
test2.twig 1            [foo]   []      [baz]
test2template.twig 1    []      []      [baz]
test2.twig 2            []      []      [baz]
test2template.twig 2    []      []      [baz]
test2.twig 3            [foo]   []      [baz]
test2page.twig 2        [foo]   []
test2layout.twig 2      [foo]   [bar]
```

The variable foo is set to 'foo' as part of the render function and the way I understand it, it should be available everywhere (I'm not doing include with "only" at any point). But it's not available in the block `test2.twig` embeds within `test2template.twig`, or in the non-block parts of `test2template.twig` embedded in `test2.twig`. (This is the main problem.)

bar is set to 'bar' in the layout which `test2page.twig` extends, and I'm not sure whether that should then be available in `test2page.twig` and the file that this one includes. I would have thought it should, but it is not.

baz is set to `baz` in `test2.twig`, which is included by `test2page.twig`. It's then available everywhere I'd expect.

Is this indeed a bug, or am I misunderstanding how things are supposed to work?

evgenius:
@tremby There may be a bug, thank you for posting this issue.
The way the context is handled is different compared to Twig.php so the more cases and tests we have, the closer its behavior will be.

dotmagic:
14 months later i run into the same issue :(
My workaround was to use `include` and pass data with `with` instead `embed`.
Actually the embed has a very limited use.

smutnyleszek:
Same problem here. It basically boiled down to this:

```twig
{% set foo = '123' %}

{% embed 'test.html.twig' with {
    bar: '456'
} %}
    {% block attributes %}
        {{ foo }} {# <-- here foo is undefined #}
    {% endblock attributes %}
{% endembed %}
```

My fix was to pass `foo` with `with` object

olets:
@smutnyleszek what version of twig.js are you using, and how are you using it?

smutnyleszek:
@olets twig version is `1.10.5` and I run it via `grunt` with this options:

```
const template = Twig.twig({
    strict_variables: true,
    id: templateName,
[...]

PR #550, opened 2018-05-02
Extract parse state

willrowe:
- This is a fairly large PR since the state while parsing a template touches a lot of the code. I have broken it down into smaller commits to hopefully make it easier to follow.
- The reason for this change was discovered while re-implementing how blocks are rendered. I realized that a lot of the issues were due to the fact that the parse "state" was being stored on the template itself. This makes it harder to manage that state and was leading to it being mishandled in multiple places. Especially when the same template was being rendered more than once by being embedded, and even more so when a template was rendering itself, within itself.
- By extracting all the code that pertains specifically to generic parsing of tokens, the state can be nested, initialized, and stored depending on the case that is required internally. Another benefit of this approach is that, depending on conditionals in the template which can change each time a template is parsed, the blocks and macros can be different; which was not necessarily taken into account before.
- The items that were moved to `Twig.ParseState` were: `blocks`, `context` (global parse context), `extend`, `importedBlocks`, `Twig.importBlocks()` `macros`, `nestingStack`, `originalBlockTokens`, `Twig.parseAsync()`, and `Twig.parse()`. A reference to the template related to the tokens being parsed is also stored on it.
- Calls to logic, expression, filter, and function `parse` methods are now bound to `Twig.ParseState` instead of `Twig.Template`. This represents the biggest user-facing change and is a __BREAKING CHANGE__. I believe this would require a major version bump and a note in the docs. Any third-party code that adds functions, filters, tags, or other extensions to the library would need to be updated to properly reference `Twig.ParseState` and `Twig.Template` (which could be accessed as a property on `Twig.ParseState`).
- This is also a __BREAKING CHANGE__ for any third-party code that call the `Twig.parse` and `Twig.parseAsync` functions directly, since they are now on the `Twig.ParseState` prototype.
- This fixes both #287 and #352, including tests for those cases.
- This the last PR needed to prepare for re-implementing the block parsing. I do not believe there will be any breaking changes in that PR.

### DIFF
--- a/src/twig.async.js
+++ b/src/twig.async.js
@@ -8,8 +8,8 @@ module.exports = function (Twig) {
     var STATE_RESOLVED = 1;
     var STATE_REJECTED = 2;
 
-    Twig.parseAsync = function (tokens, context, blocks) {
-        return Twig.parse.call(this, tokens, context, true, blocks);
+    Twig.ParseState.prototype.parseAsync = function (tokens, context, blocks) {
+        return this.parse(tokens, context, true, blocks);
     }
 
     Twig.expression.parseAsync = function (tokens, context, tokens_are_parameters) {
--- a/src/twig.async.js
+++ b/src/twig.async.js
@@ -17,7 +17,9 @@ module.exports = function (Twig) {
     }
 
     Twig.logic.parseAsync = function (token, context, chain) {
-        return Twig.logic.parse.call(this, token, context, chain, true);
+        var state = this;
+
+        return Twig.logic.parse.call(state.template, token, context, chain, true);
     }
 
     Twig.Template.prototype.renderAsync = function (context, params) {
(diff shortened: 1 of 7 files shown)


willrowe:
- @dave-irvine would it be possible to start a `2.x` or `develop` branch that could hold this and the other needed PRs with big changes until v2.0.0 itself is ready to be released?
- I should have more time starting next week to push things forward.

dave-irvine:
@willrowe done, do you want me to merge this to the develop branch?

willrowe:
@dave-irvine I'll give it a good once over and then I'll let you know if it's ready to merge.

willrowe:
@dave-irvine I just went over it very carefully and everything looks good to go. You can merge it whenever you're ready.

dave-irvine:
@willrowe are we just merging this to develop or are we going for a master release?

willrowe:
Just into develop for now, it looks like I have two other PRs that I would like to get in before doing a v2 release. We'll have to merge in the changes from the master, but I'll be happy to put together a PR for that as well at the time.

PR #556, opened 2018-05-18
improve namespaces: support non namespaces path together with namespa…

### LINKED ISSUE #555: namespaces do not work well
I'm using `metalsmith-twig` (wrapper for `twig.js`) for a project and have the following difficulties implementing namespaces.

- `metalsmith-twig` does not support namespaces (so I will make a PR soon)
- When I add support for namespaces to `metalsmith-twig` (locally) my project is broken. Current relative paths (like `_base.twig`) can't be resolved anymore.
- I think I've found the problem in `Twig.path.parsePath` (file: `twig.path.js`), because it returns `file` (string) instead passing it through the `relativePath` function. Probably this works when you're **always** using namespaces for paths and configure the namespace path absolute.

But i think it might be usefull you can use both situations together, so pass always the path string through the `relativePath` function.

```
{# Situation 1: #}
  {% extend '_base.twig' %}

{# Situation 2: #}
  {% include 'namespace::_example.twig' %}
```

So i will suggest to remove line 40 (`return file;`).

boschr:
# Improve namespaces

add support non namespaces path together with namespaces configuration

Fixes #555

### DIFF
--- a/src/twig.path.js
+++ b/src/twig.path.js
@@ -31,13 +31,14 @@ module.exports = function (Twig) {
 
         if (hasNamespaces){
             for (k in namespaces) {
-                if (colon.test(file))
+                if (colon.test(file)) {
                     file = file.replace(k + '::', namespaces[k]);
-                else if(atSign.test(file))
+                    return file;
+                } else if (atSign.test(file)) {
                     file = file.replace('@' + k, namespaces[k]);
+                    return file;
+                }
             }
-
-            return file;
         }
 
         return Twig.path.relativePath(template, file);
--- a/test/templates/namespaces_without_namespace.twig
+++ b/test/templates/namespaces_without_namespace.twig
@@ -1,0 +1,3 @@
+{% include "namespaces/namespaces.twig"  %}
+
+{% include "test::namespaces.twig" %}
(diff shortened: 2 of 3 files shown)


willrowe:
Please add a test to cover the issue you were having in #555.

boschr:
@willrowe fixed your comments and add an `.editorconfig` for more indenting consistency. We also should consider an js/es linter (like `eslint`).

boschr:
@willrowe can you confirm my changes? In case you're waiting for a new commit, i did not, just amend or fixup'ed the rework. Would be really nice if we can merge this soon. Thnx!

willrowe:
- Please remove the `.editorconfig` file, this and linting would be helpful, but I don't want to add anything extra to this PR.
- After that, should be good to go.

boschr:
@willrowe @dave-irvine i already revert the commit where i added the `.editorconfig`. Can we merge this one?

willrowe:
Looks good to me.

boschr:
Thnx guys! Who can release an new version? Would be very nice.

boschr:
@willrowe @dave-irvine any idea when we can release a new version?

rutgerbakker95:
@willrowe @dave-irvine could you guys please release a new version? Thx!!!

dave-irvine:
I'll look into this tonight. It'd be great if someone could work on some CI/CD to make the task easier, right now its a real pain to generate releases!

dave-irvine:
@boschr @rutgerbakker95 1.12.0 released

PR #559, opened 2018-05-24
Fix extend when add is null

RobLoach:
When `add` is null, there would be an exception thrown. Having it force an empty object when add isn't there fixes the issue.


dave-irvine:
Hi Rob,

Can you add a test that would fail before your fix and passes after your fix? This helps us prevent regressions.

Thanks!

RobLoach:
@dave-irvine There we go! Let me know if there's any other way to describe it better.

Issue #463, opened 2017-03-19
Test result is dependent on the name of the test file


ericmorand:
While I was trying to fix an issue, I discovered that for a same test, the result differs depending on the name of the file that implements the test.

For example, consider the following test:

```
var Twig = Twig || require("../twig"),
    twig = twig || Twig.twig;

describe("Twig.js TEST ->", function () {
    it("should render", function () {
        twig({
            path: 'test/templates/include.twig',
            async: false
        }).render({test: 'tst'}).should.equal("BeforeTest template = tst\n\nAfter");
    });
});
```

Implement this test in the file `test/test.aaa.js` and it will pass.
Implement this test in the file `test/test.zzz.js` and it will fail.

Why? Because the same instances of Twig is shared between all the tests and never _reset_. Thus, any change made to the Twig instance is propagated to the subsequent tests, leading to a **totally unreliable tests suite**! **A test that pass with success doesn't guarantee that the same exact piece of code will execute with success in another project**.

That's a pretty critical issue that defeat the purpose of the whole test suite of this project.

dave-irvine:
Well. That sucks!

Thanks for noticing it.

I guess we need to review the test suite and add a `before()` to every test set where we reset/re-import the Twig instance.

ericmorand:
Yes, probably.

I ended up deleting the require cache for twig.js as you can see [here](https://github.com/ericmorand/twig.js/blob/edb15e29c90cb1502e213fd4c4b7ceb65ba023d9/test/test.rethrow.js#L3) and it did the trick.

dave-irvine:
Ok well that would be a fairly fast hack-fix to implement. Do you want to give it a try and see how much of the test suite fails? :D

ericmorand:
Yep, I would gladly do that.

I will also try to have Twig being a constructor, thus having `let twig = new Twig();` instead of `let twig = Twig.twig`; and see how everything explode 😄 . That would be the best approach to the problem: we could keep require cache and still have a fresh twig instance. That would obviously be a breaking change so consider this as a proof-of-concept.

ericmorand:
@dave-irvine, I have great news: only **two** tests fail when applying the hack-fix, both in `test.macro.js`. And they fail because they were wrong in the first place.

Take a look at this one:

```
it("should support inline includes by ID", function() {
        twig({
            id:   'hello',
            data: '{% macro echo(name) %}Hello {{ name }}{% endmacro %}'
        });

        var template = twig({
[...]

PR #561, opened 2018-05-31
add a factory function to get a completely fresh Twig object

GertSallaerts:
This adds a `factory` function to the exported Twig object. By calling the `factory` method, one can obtain a fresh Twig object. This means all custom filters, functions, ... will be gone and the developer can start fresh.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -604,21 +604,6 @@
         "unset-value": "^1.0.0"
       }
     },
-    "caller-path": {
-      "version": "0.1.0",
-      "resolved": "https://registry.npmjs.org/caller-path/-/caller-path-0.1.0.tgz",
-      "integrity": "sha1-lAhe9jWB7NPaqSREqP6U6CV3dR8=",
-      "dev": true,
-      "requires": {
-        "callsites": "^0.2.0"
-      }
-    },
-    "callsites": {
-      "version": "0.2.0",
-      "resolved": "https://registry.npmjs.org/callsites/-/callsites-0.2.0.tgz",
-      "integrity": "sha1-r6uWJikQp/M8GaV3WCXGnzTjUMo=",
-      "dev": true
-    },
     "camelcase": {
       "version": "4.1.0",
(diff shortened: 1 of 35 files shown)


willrowe:
This is interesting, can you add some tests to ensure it is working as expected?

ericmorand:
This may come as an elegant solution to #463!

GertSallaerts:
I'll try to get to it this weekend.

RE #463 I could indeed remove the "re-require" hack and rewrite all the tests to use my factory instead if you like, should produce the same result.

ericmorand:
Yes, would be awesome. Actually fixing the issue and remove the require-uncached kind of hack.

GertSallaerts:
Added tests for the new functionality and removed the old require-uncached solution.

dave-irvine:
@GertSallaerts I broke your pull, I'm sorry :( Can you rebase and try again?

GertSallaerts:
Rebased @dave-irvine

dave-irvine:
@GertSallaerts thanks. I'll build a new release in a couple of weeks when we get some more pulls in.

willrowe:
This has somehow broken the test suite. It is [only running the two tests added by this PR](https://travis-ci.org/twigjs/twig.js/builds/390977346#L954-L959) for some reason.
@GertSallaerts could you please take a look and see if you can figure out what is causing the other tests to not be run?

dave-irvine:
@willrowe Good catch.

I think its `describe.only()` causing the issue, can you check @GertSallaerts ?

willrowe:
I just tested it without `only` and that seems to be the issue.
It was probably added to only test those new tests.
I also just noticed that `const` is being used [here](https://github.com/twigjs/twig.js/blob/13cd136b747567c18493d785eb7931c29285ef04/test/test.factory.js#L15) and [here](https://github.com/twigjs/twig.js/blob/13cd136b747567c18493d785eb7931c29285ef04/test/test.factory.js#L31), it works no problem since it's just in tests, but `const` is not used elsewhere so not sure if that is a problem at all.

GertSallaerts:
I will make an extra PR to clean it up

Issue #567, opened 2018-06-21
Multiple namespaces not working


Regaez:
You seem to be only able to use one namespace... it'd be nice if you could configure multiple.

dave-irvine:
Resolved by https://github.com/twigjs/twig.js/pull/568

PR #568, opened 2018-06-21
Allow for multiple namespace resolutions

Regaez:
Addresses #567 

Changes the namespace regex to take the namespace key into account, so that multiple namespaces can be resolved.


dave-irvine:
This is the kind of PR I like, one that addresses the issue the puller raised :) Thanks!

Issue #571, opened 2018-06-29
RAW filter didn't working!!!


IftekherSunny:
Any reason doesn't support for RAW filter???

dave-irvine:
The reason is it hasn't been implemented yet.

IftekherSunny:
implement it bro :(

dave-irvine:
I look forward to your pull request :)

toptalo:
As I see templates rendered as raw by default, because template block `{% autoescape %}` hasn't been implemented yet. Do you need it, or what you use case?

maybe you need `{% verbatim %}` block, that are no implemented too

toptalo:
Hmmmm
I found some strange thing with `raw` and `verbatim` tags.

At first `{% raw %}`  tag was defined in `twig.core.js` and on commit 6aa9e97 Helder Santana has rename `{% raw %}` to `{% verbatim %}` tag. But after merge ac631f0 this change was lost and later returned in 9f0435b.

So for now, with my help there are two places where `{% verbatim %}` tag has defined: `twig.core.js` and `twig.logic.js`

Also we have `{% raw %}` tag that are outdated.

---

I think we need to revert my changes and also remove outated `{% raw %}` tag

PR #574, opened 2018-07-03
Add `verbatim` tag

toptalo:
Address issue #571

The `verbatim` tag marks sections as being raw text that should not be parsed. For example to put Twig syntax as example into a template you can use this snippet:
```
{% verbatim %}
    <ul>
    {% for item in seq %}
        <li>{{ item }}</li>
    {% endfor %}
    </ul>
{% endverbatim %}
```


toptalo:
Thanks!

PR #580, opened 2018-08-13
multiple namespaces fix

stefspakman:
Fix to allow multiple namespaces

### DIFF
--- a/src/twig.path.js
+++ b/src/twig.path.js
@@ -27,6 +27,9 @@ module.exports = function (Twig) {
 
         if (hasNamespaces){
             for (k in namespaces) {
+              if (file.indexOf(k) === -1) {
+                continue
+              }
 
                 // check if keyed namespace exists at path's start
                 var colon = new RegExp(`^${k}::`);
--- a/src/twig.path.js
+++ b/src/twig.path.js
@@ -27,9 +27,9 @@ module.exports = function (Twig) {
 
         if (hasNamespaces){
             for (k in namespaces) {
-              if (file.indexOf(k) === -1) {
-                continue
-              }
+               if (file.indexOf(k) === -1) {
+                  continue
+               }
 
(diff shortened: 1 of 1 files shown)


dave-irvine:
Hi @stefanullinger , thanks for your contribution. Could you add a unit test that fails before your change and passes afterwards?

stefspakman:
Hi @dave-irvine, 
I just tried making a testcase for this fix, however the current multiple namespaces test already seems to be passing every time. As i do not have any experience in writing unit tests, i'm not sure how to best recreate my case as a test.

The problem I had were as follows:
when i declared multiple namespaces, only the first one would actually work, any other would not be picked up by the render function. 

Twig.twig({
        id: id,
        data: twig,
        namespaces: {
        'atoms': path.resolve(__dirname, `${process.cwd()}/patterns/atoms/`),
        'molecules': path.resolve(__dirname, `${process.cwd()}/patterns/molecules/`)
      }
      }).render(json)

when my twig uses {% include "@molecules/testfile.twig"%}, it won't find the file because it looks at /patterns/atoms/testfile.twig, instead of patterns/molecules/testfile.twig

RobLoach:
Seems like this is more of a performance update? It says if the namespace is not found in the given file, skip it. In theory, tests should pass with or without this change.

src/twig.path.js
```diff
@@ -27,6 +27,9 @@ module.exports = function (Twig) {
 
         if (hasNamespaces){
             for (k in namespaces) {
+              if (file.indexOf(k) === -1) {
+                continue
+              }
```
RobLoach: Should be 3 spaces instead of 2.

stefspakman:
Hi @RobLoach,
Yes, maybe I should have classified it as an performance update. The issue occurred when processing a lot of files, with this change it should process the namespaces faster.

Issue #578, opened 2018-08-07
Using 'matches' comparison operator brings "Fatal error: Cannot read property 'length' of undefined"


hdwebpros:
I'm using this
 ```
                {#  This breaks it.
                    I'm trying to detect if the first couple of characters are numbers, like 00, 01, etc.
                    https://twig.symfony.com/doc/2.x/templates.html#comparisons
                    The 'matches' operator seems to break things here
                #}
                {% if variable[:2] matches '/^\\d+$/' %}
                    {% set variable = stripped[:3] %}
                {% endif %}
```
The example variable I tried with was `00-test`.

I got 
"TwigException: Twig.expression.type.variable cannot follow a Twig.expression.type.slice at template:12 near 'matches...'
    Fatal error: Cannot read property 'length' of undefined"

willrowe:
As far as I know the `matches` comparison operator is not supported by this library yet.

willrowe:
This is a duplicate of #505.

Issue #505, opened 2017-11-03
[ Feature ] Missing 'matches' expression


julianxhokaxhiu:
Hi,

I was using this awesome library, and i am really surprised how close is to Twig for PHP. Although I found that one expression is missing, which I think is very useful to have: `matches`.

You can see more about it here: https://twig.symfony.com/doc/2.x/templates.html#comparisons

Do you think is it possible to have it on Twig.js?

Best regards,
Julian

dave-irvine:
Absolutely possible, just need someone to write it! :)

julianxhokaxhiu:
I'll see if I will find some time tomorrow and I will be able to forge a PR then. Thanks!

mortenson:
Looks like `starts with` and `ends with` are missing too - I found this old commit to add them if it helps out at all: https://github.com/twigjs/twig.js/commit/474d73639f790dead4f6010583a6dcdfc3566fd6

dave-irvine:
@mortenson good find. I think someone just needs to cherry-pick that commit id to a new branch and create a new PR.

willrowe:
This is a duplicate of #381.

Issue #381, opened 2016-06-01
`matches` is not supported


lvl99:
RegExp matching using the `matches` expression doesn't seem to be supported:

```
var Twig = require('Twig')
var twig = Twig.twig
var testTemplate = twig({
  data: "{{ var }} is {% if var matches '/^\\d+$/' %}a number{% else %}not a number{% endif %}"
})
console.log(template.render({
  var: "5"
}))
```

I looked at the implementation notes and it seems there's no mention if `matches` is supported or not, so I assume it isn't?

Here's a twigfiddle showing the expected results with Twig PHP: http://twigfiddle.com/zm6rcq

dave11mj:
I am experiencing the issue

`Twig.expression.type.variable cannot follow a Twig.expression.type.variable at template:20 near 'matches...'`

PR #586, opened 2018-09-06
Add more comparison operators

toptalo:
Add `starts with` and `ends with` comparison operators and also the `matches` operator that allows you to use regular expressions

Address #381 #505 #578


toptalo:
Open new PR #587

PR #588, opened 2018-09-07
Filter url_encode was extended for objects

ahtohbi4:
The ability to encode an object to URL by the filter `url_encode` was added according to [the Twig PHP](https://twig.symfony.com/doc/2.x/filters/url_encode.html#url-encode).

```javascript
const data = {
    foo: {
        baz: 'bar',
    },
};
const template = twig({ data: '{{ foo|url_encode }}' });

console.log(template.render(data)); // => "foo%5Bbaz%5D=bar"
```

### DIFF
--- a/src/twig.filters.js
+++ b/src/twig.filters.js
@@ -138,6 +138,30 @@ module.exports = function (Twig) {
                 return;
             }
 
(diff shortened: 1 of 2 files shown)


ahtohbi4:
What about this PR? Are there some doubts?

PR #589, opened 2018-09-15
WIP: esling and pre-commit hook were added

ahtohbi4:
I have thought that it will be a good idea to use `eslint` to normalize a code style.

Also, I have added `.editorconfig` and pre-commit hook to automatic fix code from contributors.

What do you think?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+# EditorConfig is awesome: https://EditorConfig.org
+root = true
+
+[*]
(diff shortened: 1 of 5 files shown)


dave-irvine:
So eslint is fine, thanks for adding that, however .editorconfig I would rather we use Prettier with eslint integration, is that OK?

ahtohbi4:
Sounds good. I'll think about it.

Issue #594, opened 2018-11-02
Provide un-minified sources in the npm package


dimd13:
When I write code, I, like many others, have a habit of looking into the source code of a module to see how it works, what arguments it takes, and so on. But the source code (twig.js file) in the package from NPM is minified, which makes it very difficult to read. Please fix it

dave-irvine:
Thanks for the suggestion @dimd13, I'm looking forward to your pull request :)

toptalo:
We can close this issue

RobLoach:
Thanks @toptalo !

PR #598, opened 2018-12-07
Change webpack config to provide un-minified sources in the npm package

toptalo:
After build only `twig.min.js` will be minimized

Address issue #594


RobLoach:
Thank you for putting this together! 👍

PR #604, opened 2019-01-19
fix for not autoescaping includes

tgabi333:
The previous autoescaping improvement (https://github.com/twigjs/twig.js/pull/577) broke includes.


RobLoach:
Thank you, was wondering where things broke.

Issue #608, opened 2019-03-11
Namespaces: Unable to find template file due to missing path separator


radiocity:
Twigjs throws `TwigException: Unable to find template file C:\Projects\test\src\layoutscommon.twig`
```js
// ...
namespaces: {
  'layouts': path.join(__dirname, '/src/layouts'),
},
// ...
```
```twig
{% extends "layouts::common.twig" %}
{% block content %}Lorem ipsum{% endblock %}
```

Everything works if separator added
```js
// ...
namespaces: {
  'layouts': path.join(__dirname, '/src/layouts/'),
},
// ...
```

willrowe:
This was fixed by #609. Example: https://codepen.io/willrowe/pen/YzarBQP

PR #609, opened 2019-03-26
Add possibility to define namespace without slash at the end of the path

toptalo:
Address #608


RobLoach:
Thanks a lot for the updates here. Glad some more attention has been put into the namespace/pathing system.

PR #619, opened 2019-05-05
Replace UglifyJs with Terser as the webpack minimizer

PolyPik:
Terser can handle ES6 syntax unlike UglifyJS.


RobLoach:
Thanks for these updates! :+1:

PR #625, opened 2019-05-11
Merge block re-implementation for release

willrowe:
This brings the `develop` branch inline with the `master` branch and cleanly merges it into `master`.

This means that the block re-implementation is ready for release. This will require a v2.0.0 release because of the BC breaks.

@RobLoach you can do a pre-release version for v2.0.0 if you want to let people test before release, up to you. It might be nice though, I'd be glad to go through all the fixed issues and add a note about the pre-release version so those who submitted the specific issues can be test against their real-world usages.

### DIFF
--- a/src/twig.async.js
+++ b/src/twig.async.js
@@ -8,16 +8,20 @@ module.exports = function (Twig) {
     var STATE_RESOLVED = 1;
     var STATE_REJECTED = 2;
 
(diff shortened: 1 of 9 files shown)


PolyPik:
@RobLoach Please merge this first before you merge my PR (i.e. #626) just to be on the safe side.

PR #628, opened 2019-05-15
XO Updates

PolyPik:
Prior Travis builds were failing due to `package-lock.json` not including `xo`. I took this as an opportunity to not just include the dependency in `package-lock.json` but to also regenerate the lock file from scratch.

### DIFF
--- a/lib/compile.js
+++ b/lib/compile.js
@@ -1,96 +1,102 @@
-var Twig = require("../twig")
-    , twig = Twig.twig
-    , PATHS = require("./paths")
-    , MINIMATCH = require("minimatch")
-    , FS = require("fs")
-    , WALK = require("walk");
+const FS = require('fs');
+const minimatch = require('minimatch');
+const WALK = require('walk');
+const Twig = require('../twig');
+const PATHS = require('./paths');
+
+const {twig} = Twig;
 
 exports.defaults = {
-    compress: false
-    , pattern: "*\\.twig"
-    , recursive: false
+    compress: false,
+    pattern: '*\\.twig',
+    recursive: false
 };
 
-exports.compile = function(options, files) {
+exports.compile = function (options, files) {
     // Create output template directory if necessary
     if (options.output) {
         PATHS.mkdir(options.output);
     }
-    
-    files.forEach(function(file) {
-        FS.stat(file, function(err, stats) {
+
+    files.forEach(file => {
+        FS.stat(file, (err, stats) => {
             if (stats.isDirectory()) {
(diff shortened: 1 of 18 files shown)


PolyPik:
Okay the

PolyPik:
Okay the tests can now run again, so the previous issue is resolved. But now that xo is working the builds are failing due to the many style and format violations.

@RobLoach should we enforce camelcase?

RobLoach:
All that xo recommends is valid, if we want to adopt xo for syntax checking. I believe our webpack build brings things back to es5 for browser compatability.... Right?

PolyPik:
Webpack by itself doesn't offer transpilation.

For transpiling to ES5, we will need to use the [babel-loader](https://github.com/babel/babel-loader).

PolyPik:
These commits will de-linted the files checked by XO, but only to the extent that errors are no longer being given off. The files still have several warnings that have yet to be resolved.

willrowe:
This needs to be reverted. Usage of `const` and `let` cannot be included until we have some kind of transpiling setup. See #471 where this has been addressed this before.

I'm not very clear on what exactly this PR was originally meant to do, can you explain @PolyPik?

PolyPik:
This PR was meant to add a de-linter to the project to help enforce a consistent coding style and format.

willrowe:
Thanks for clearing that up. Seeing as this breaks browser compatibility, we should probably just remove the `xo` check in `posttest` for right now in addition to reverting this PR so that Travis will not fail. What do you think @RobLoach?

PolyPik:
Is there any reason we couldn't just add babel-loader to our webpack.config?

If the issue is that ES6 code is not being transpiled, then we should add a transpiler instead throwing away these changes.

willrowe:
I would rather have the transpiler in place before making these changes and not have to rush in choosing one right now. This could possibly hold up the v2.0.0 release as well, so since it's just code style and not changing any functionality or fixing any bugs, I would like to revert and revisit later.

PolyPik:
So will this revert go about? Will it involve `git push --force` or will it involve making a new commit that simply sets the state of the repo to that of an older commit?

RobLoach:
Could we move the transpiler in place? May be less work to add the transpiler than revert, and re-revert the change.

willrowe:
If someone has time to put together a PR for adding a transpiler, that would work. The order of things is not ideal, but at this point it's probably fine since no release has been made since these changes.

Sorry for jumping to revert so quickly, I kind of panicked when I saw the changes since I missed this PR initially.

PR #630, opened 2019-05-16
Update Mocha and fix Twig.forEach

RobLoach:
@PolyPik Unsure of the correct solution around this one. Do you know?

### DIFF
--- a/src/twig.core.js
+++ b/src/twig.core.js
@@ -57,14 +57,14 @@ module.exports = function (Twig) {
     }
 
     Twig.forEach = function (arr, callback, thisArg) {
-        if (Array.prototype.forEach ) {
+        if (Array.isArray(arr)) {
             return arr.forEach(callback, thisArg);
         }
 
         var T, k;
 
         if ( arr == null ) {
-          throw new TypeError( " this is null or not defined" );
+          return;
         }
 
         // 1. Let O be the result of calling ToObject passing the |this| value as the argument.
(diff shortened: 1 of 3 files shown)


src/twig.core.js
```diff
@@ -57,14 +57,14 @@ module.exports = function (Twig) {
...
-        if (Array.prototype.forEach ) {
+        if (Array.isArray(arr)) {
             return arr.forEach(callback, thisArg);
         }
 
         var T, k;
 
         if ( arr == null ) {
-          throw new TypeError( " this is null or not defined" );
+          return;
```
RobLoach: Unsure this is the correct way around this.

src/twig.core.js
```diff
@@ -57,14 +57,14 @@ module.exports = function (Twig) {
     }
 
     Twig.forEach = function (arr, callback, thisArg) {
-        if (Array.prototype.forEach ) {
+        if (Array.isArray(arr)) {
             return arr.forEach(callback, thisArg);
```
RobLoach: Should this allow looping through keys on an object?

PolyPik:
Twig.forEach and Twig.indexOf are polyfills for IE8. I doubt we will be supporting such an old browser.

Internet Explorer has been able to do all of the Array methods since IE9.

https://kangax.github.io/compat-table/es5/

My suggestion replace these polyfills with proper calls to the appropriate Array method.

PR #637, opened 2019-05-26
add json_decode filter

eistrati:


RobLoach:
Thanks for the addition. Would you mind providing some tests to ensure it's longevity?

PolyPik:
json_decode is not feature found in the reference implementation (i.e. TwigPHP). Should we include a non-standard language feature?

eistrati:
Gentlemen,

I was trying to load JSON from file into TWIG variable, but it was useless without `json_decode` filter. My code (inspired from this link - https://stackoverflow.com/questions/44654948/set-twig-variable-to-json-file-as-an-include) looked like this:

```
{% set recipes %}
  {% include 'metadata.json' %}
{% endset %}

{% for key,value in metadata %}
  {{ key }} => {{ value }}
{% endfor %}
```

After adding `json_decode` filter next to `json_encode`, my code looks like this (AND IT WORKS!!!):

```
{% set recipes %}
  {% include 'metadata.json' %}
{% endset %}

{% for key,value in metadata|json_decode %}
  {{ key }} => {{ value }}
{% endfor %}
```

Now, I have solved my problem and could just point to https://github.com/MitocGroup/twig.js in package.json, but I thought it would be helpful for this feature to be part of https://github.com/twigjs/twig.js. Please help me by pointing out to `json_encode` tests, and I'd be happy to create `json_decode` tests as well. Nevertheless, you decide if you want it part of your codebase.

Thank you,
Eugene.

PolyPik:
Is there any reason why you couldn't have loaded the json into a regular Javascript object then pass that object to the template via the render context?

eistrati:
Yes, there is. We use TWIG for local development, while during deployment we generate static HTML and CSS that are pushed to S3 + CloudFront on top of it. That means that JSON is not used at runtime, but instead it is used during build and deploy steps. Since this is just a simple CMS-like solution, we believe there is no need to add this overhead at runtime. I hope that answers your question...

PolyPik:
Hmm, in that case if this feature were to be added we should note in the documentation that `json_decode` is **not** a standard language feature so that template developers can choose to use this filter at their own discretion.

toptalo:
There are possibility to add a custom filters why you want use it?

eistrati:
Honestly, I believe that `json_decode` filter should be part of basic codebase next to `json_encode` filter, but at the end of the day it's your call and you guys should tell me if you decide to make it part of the core functionality or not. Thank you!

willrowe:
I would rather not add something that is not available in the PHP version that this library is based on, especially when it comes to filters, etc. It would need to have a pretty strong JavaScript-only use case.

It looks like it was proposed [here](https://github.com/twigphp/Twig/pull/824), where you can see their rationale for not adding it in.

RobLoach:
@eistrati It could live as an addon package over at [kalastatic-twig-filters](https://github.com/kalamuna/kalastatic-twig-filters).

Issue #612, opened 2019-04-30
[PROPOSAL] Make all loaders asynchronous-only and Promise-based.


PolyPik:
@RobLoach 
What I am proposing will introduce breaking-changes, thus before I even make a draft PR, I wish to discuss this in Issues.

Currently, the engine provides the user the option of choosing to load templates synchronously or asynchronously. This makes sense for the file loader as Node provides synchronous and asynchronous ways to read files on the filesystem. However, for ajax the option of having it perform synchronously defeats the entire point of ajax which is suppose to be inherently asynchronous.

At first, I've tried to make the ajax loader async-only and allow for sync and async loading for the file loader. But I then encountered issues when I tried to ajax load templates that use other templates (i.e. include, extends, embed, import). The code that is suppose to handle these associations work on the assumption that the loader being used is able to perform a synchronous load.

Given what I've learned, I propose that the code be refactored in a way that assumes that all loaders are asynchronous-only. If the user wants to synchronously load a template, that can be achieved via the `data` key when providing params to the engine.

In addition, given that the project supports a minimum Node version of 6 (as indicated by the .travis.yml), I also propose that we deprecate the current use of callbacks to implement asynchronous behavior in favor a Promise-based implementation. Doing so would also allow for the server, in addition to the browser, to perform ajax loading through the use of Promise-based http clients such as [axios](https://www.npmjs.com/package/axios).

PolyPik:
I've already discovered this a few weeks before, but now I want to bring it to everyone's attention.

In the process of changing all the loaders to async only, I discovered that the change would also result in template rendering also becoming async only. The reason for this is because the action of rendering can involve template loading, due to the use of tags like `include`, `extends`, and `embed`, in the initial template.

If template loading happens asynchronously, then template rendering must happen asynchronously as well. Given that making all loaders asynchronous only would result in a breaking change to the engine's public API, should we still proceed with this course of action?

PolyPik:
@dave-irvine, @RobLoach, @willrowe 

Should we proceed?

willrowe:
I have re-written the load a couple of times to be async-only before the hybrid implementation was done a while back, so I'm definitely on board for loading and rendering being async-only.

I wasn't involved in the hybrid implementation, so I'm not sure why it was done that way or if there is anything to take into account for that. I can look over the old issues and PRs for it though and double-check if that would be helpful.

PolyPik:
[...]

PR #638, opened 2019-05-30
Make loading and rendering async only

PolyPik:
This commit will make loading a template from a file, local or
remote, asynchronous only.

The only way to have synchronous template loading is to provide a string or
array of compiled tokens to the data attribute of the parameters object
passed to `Twig.twig`.

Addresses #612


willrowe:
I can take a look at this next week, a bit swamped this week.

RobLoach:
This is some incredible change. Nicely done... Are we changing some of the APIs?

Will the build process adapt this for browsers?

PolyPik:
> This is some incredible change. Nicely done... Are we changing some of the APIs?
> 
> Will the build process adapt this for browsers?

This commit makes rendering a purely asynchronous process, so yes there are public API changes.

As for the build process, all it does right now for the browser build is create a minified version of the server build. I think for an upcoming task we can do a bit better by excluding the fs loader entirely from the browser build.

PolyPik:
@dave-irvine Are these changes acceptable?

willrowe:
@PolyPik I am planning on taking a look over this tomorrow if you would still like me to.

PolyPik:
> @PolyPik I am planning on taking a look over this tomorrow if you would still like me to.

I would like that very much. 

My concern is that I mistranslated certain tests and that the CLI not longer works.

willrowe:
> My concern is that I mistranslated certain tests and that the CLI not longer works.
I'll be sure to double-check that.

willrowe:
I started taking a look over this, but then realized that #628 needs to be reverted.

There's also a lot of issues involved with too many changes being made at once that I'm afraid will cause things to break. I think we need to go back to the drawing board on this and try again with incremental changes instead of one large PR.

A good, simple start would be to just remove the option to render synchronously, then all synchronous-specific code could be removed.

PolyPik:
Closing this PR. These changes will be a part of version 2.0.0.

PR #639, opened 2019-06-05
Add Babel

RobLoach:
This brings in Babel to the webpack build to ensure the browser compatibility is kept in place.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "posttest": "xo src lib bin"
   },
   "dependencies": {
+    "@babel/runtime": "^7.4.5",
     "locutus": "^2.0.5",
     "minimatch": "3.0.x",
     "walk": "2.3.x"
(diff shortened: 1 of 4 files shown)


webpack.config.js
```diff
@@ -11,6 +11,27 @@ module.exports = {
...
+            {
+                exclude: /(node_modules)/,
+                use: {
+                    loader: 'babel-loader',
+                    options: {
+                        presets: ['@babel/preset-env'],
+                        plugins: [
+                            [
+                                "@babel/plugin-transform-modules-commonjs", {
+                                    "allowTopLevelThis": true
```
PolyPik: If we remove the `allowTopLevelThis: true` will the tests still pass?

RobLoach: Unsure, let's check.

RobLoach: Looks like it passed :+1:

package.json
```diff
@@ -33,11 +33,17 @@
     "posttest": "xo src lib bin"
   },
   "dependencies": {
+    "@babel/runtime": "^7.4.5",
```
RobLoach: Is babel runtime needed as a normy dependency? What does this do? Does it change how anything works?

PolyPik: The use of `@babel/runtime` and `@babel/plugin-transform-runtime` is suppose to cut down on potential code bloat. Without `@babel/runtime`, certain injected functions might be defined repeated throughout the code thus resulting in an increased bundle size. `@babel/runtime` makes so that those functions are imported only once through the bundle.

Here is an [example](https://babeljs.io/docs/en/babel-runtime) of how `@babel/runtime` is used.

RobLoach: Thanks for the clarification :+1:

PR #642, opened 2019-06-08
Streamline Webpack builds

PolyPik:
Instead of creating the server bundle and the browser bundle in two
separate Webpack builds, both bundles will be created in a single
Webpack build by having `webpack.config.js` export an array of
configuration objects.

In addition, the server build will not require transpilation as the
minimum supported NodeJS version (i.e. NodeJS 8) supports the use of
ES6 language features as is.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,10 +26,7 @@
     "postversion": "git push origin master && git push origin master --tags",
     "pretest": "npm run build",
     "test": "mocha -r should",
-    "prebuild": "npm run build-node",
-    "build-node": "webpack",
-    "build-browser": "cross-env WEBPACK_ENV=browser webpack",
-    "build": "npm run build-browser",
+    "build": "webpack",
     "posttest": "xo src lib bin"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -26,10 +26,7 @@
     "postversion": "git push origin master && git push origin master --tags",
     "pretest": "npm run build",
     "test": "mocha -r should",
-    "prebuild": "npm run build-node",
```
willrowe: > In addition, the server build will not require transpilation as the minimum supported NodeJS version (i.e. NodeJS 8) supports the use of ES6 language features as is.

Is this what you meant by the above?

willrowe:
@PolyPik I would appreciate it if at least one other contributor signed off on PRs like this before you merged it yourself.

package.json
PolyPik: Yes, that is how the server build was performed prior to this commit.

willrowe: Ok, just making sure that had nothing to do with the NPM deploy.

PR #643, opened 2019-06-08
Clean up twig.lib.js

PolyPik:
Remove `Twig.lib.isArray`, `Twig.lib.copy`, and `Twig.lib.extend`, as
those functions can be replaced by built-in language features or
functions.

Remove `test.lib.js` as it only tests the aforementioned and removed
`Twig.lib.extend`.

Update `Twig.lib.is` to use more explicit comparisons.

Update `Twig.lib.replaceAll` to use regular expressions.

Update `Twig.lib.chunkArray` to use `Array.isArray` when checking to see
if the input array is actually an Array object.


willrowe:
@PolyPik I would appreciate it if at least one other contributor signed off on PRs like this before you merged it yourself.

PolyPik:
Very well.

> @PolyPik I would appreciate it if at least one other contributor signed off on PRs like this before you merged it yourself.

Very well, then. Guess I'll have to be more patient next time.

PR #653, opened 2019-07-17
Make '??'-operator return the actual array instead of array length

diegomorales:
The '??'-operator is very handy for setting default values. Like for example:

`{% set list = list ?? [] %}`

This works in PHP Twig. And it should of course work in twig.js too :)

### DIFF
--- a/test/test.expressions.operators.js
+++ b/test/test.expressions.operators.js
@@ -73,6 +73,15 @@ describe('Twig.js Expression Operators ->', function () {
             outputT.should.equal('two');
             outputF.should.equal('two');
         });
+
+        it('should support the null-coalescing operator for true conditions on objects or arrays', function () {
+            const testTemplate = twig({data: '{% set b = a ?? "nope" %}{% for item in b %}{{item}}{% endfor %}'});
+            const outputArr = testTemplate.render({a: [1,2]});
+            const outputObj = testTemplate.render({a: {b:3, c:4}});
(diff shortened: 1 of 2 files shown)


test/test.expressions.operators.js
```diff
@@ -73,6 +73,15 @@ describe('Twig.js Expression Operators ->', function () {
...
         });
+
+        it('should support the null-coalescing operator for true conditions on objects or arrays', function () {
+            const testTemplate = twig({data: '{% set b = a ?? "nope" %}{% for item in b %}{{item}}{% endfor %}'});
+            const outputArr = testTemplate.render({a: [1,2]});
+            const outputObj = testTemplate.render({a: {b:3, c:4}});
+
+            outputArr.should.equal('12');
+            outputObj.should.equal('34');
+        });
```
RobLoach:
``` suggestion
            const outputNull = testTemplate.render();
            outputNull.should.equal('nope');
        });
```

The above test breaks, likely because of the Twig logic. But would be good to have a test that returns the false-case. Thanks!

diegomorales: Hi Rob
I just updated the test. The false case should be covered now.

RobLoach: Thanks so much, looks great :+1:

RobLoach:
Thank you! Going to merge this one. :+1:

Issue #652, opened 2019-07-03
Add filter, map and reduce features


OxCom:
There are [few nice features](https://symfony.com/blog/twig-adds-filter-map-and-reduce-features) were added to Twig.

Please add support for this features. At current stage compilation process fails on parse mode of such constructions like:
```js
{% for product in related_products|filter(product => product.stock > 10) %}
    {# ... #}
{% endfor %}
```

julmot:
I'm missing this soo much....

willrowe:
As mentioned in #861, the following needs to be done:

1. Add support for arrow functions in limited contexts to match where TwigPHP allows them.
2. Allow arrow functions to be used in filters that already exist like `sort`.
3. Add `filter`, `map`, and `reduce` filters that rely on support for arrow functions.

> Could one of the maintainers point me out into the right direction; would you like this implemented as a logic token type or as an expression type?
@antoineveldhoven which one does TwigPHP use? Trying to match that would be a good start, if possible.

JulienLavocat:
Hello,

We are currently migrating part of our existing Symfony applications to a new infrastructure and this is a blocking issue for us. I'll be glad to contribute but it would be nice if you could please give me some leads to start implementing this properly :)

PR #656, opened 2019-08-15
Add support apply tag

maxhelias:
A first step for #652


maxhelias:
@PolyPik, @willrowe Can one of you review that ? 😃

PR #661, opened 2019-09-23
Fix 'starts with' and 'ends with' expressions

### LINKED ISSUE #660: "starts with" and "ends with" expressions result in error with non-string values
## Problem/Motivation

With Twig the following is possible without getting errorr:
```
{% if some_undefined_variable starts with "foo" %}
  foo
{% endif %}
{% if some_undefined_variable ends with "bar" %}
  bar
{% endif %}
```

With Twig.js, the preceding examples will throw an error for both `starts with` and `ends with` expression:
```
TypeError: Cannot read property 'indexOf' of undefined
```

Twig.js uses `indexOf()` with `starts with` and `ends with` expressions, to check if the string starts or ends with a certain string. However, there’s no check if we are checking the start or of a string or something else. This results in an error, if the value is something else than a string.

## Proposed resolution

Add a check if `starts with` or `ends with` is run against a string value, before trying to get `indexOf()` from the value.

## Definition of done

`X starts with Y` and `X ends with Y` expressions will return `false` when `X` is a non-string value.

ilkkave:
Fixes #660

Makes `starts with` and `ends with` expressions to not throw an error with non-string values.

Changes proposed in this pull request:

- Add a check for `starts with` and `ends with` expressions to check if it’s tested against a string.
- Add a test case where using `starts with` and `ends with` with an undefined variable.

### DIFF
--- a/src/twig.expression.operator.js
+++ b/src/twig.expression.operator.js
@@ -346,11 +346,11 @@ module.exports = function (Twig) {
                 break;
 
             case 'starts with':
-                stack.push(a.indexOf(b) === 0);
+                stack.push(typeof a === 'string' && a.indexOf(b) === 0);
                 break;
 
             case 'ends with':
-                stack.push(a.indexOf(b, a.length - b.length) !== -1);
+                stack.push(typeof a === 'string' && a.indexOf(b, a.length - b.length) !== -1);
                 break;
 
             case '..':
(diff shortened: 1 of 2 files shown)


RobLoach:
Thanks, this will help clean up input exceptions...

Should this also work with numbers or values? Outside of just strings?

ilkkave:
> Should this also work with numbers or values? Outside of just strings?

I think it should only work for strings, but I can take a look how it works in Twig.

RobLoach:
This is a good change in the mean time. For array support, that should likely be a new PR. Thanks!

PR #668, opened 2019-10-18
Reusing `empty` test in `default` filter.

oleg-andreyev:
Improving `default` by reusing `empty` test.
Following logic https://github.com/twigphp/Twig/blob/55704ce1d29f184374f77b9542baaba58d4d30f2/src/Extension/CoreExtension.php#L810

### DIFF
--- a/src/twig.filters.js
+++ b/src/twig.filters.js
@@ -207,14 +207,9 @@ module.exports = function (Twig) {
 
             return output.join(joinStr);
         },
-        default(value, _default) {
-            // checking _default for empty because it's resolved as `false` in parseParams
-            if (_default === false) {
-                _default = '';
-            }
-
+        default(value, params) {
             if (Twig.tests.empty(value)) {
-                return _default;
+                return params[0] || '';
             }
 
             return value;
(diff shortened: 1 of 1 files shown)


willrowe:
Please include a test demonstrating the bug this fixes.

oleg-andreyev:
I'll need take a look when when running mocha everything is fine, using twig on web I have different results.

oleg-andreyev:
> Please include a test demonstrating the bug this fixes.

It's not longer an issue.

oleg-andreyev:
@willrowe seems that we have misunderstanding. this PR is still relevant about improving `default` filter.

but 

> It's not longer an issue.

was about {{params}} being `false` instead of `array`

willrowe:
Is this fixing some kind of inconsistency? Please explain more about what it is improving.

oleg-andreyev:
- Inconsistency 1: `default` filter in php has default value (empty), while `default` filter in js requires an argument
- Inconsistency 2: `default` filter in js has it's own logic of detecting empty value, which is not the same as `empty` test, which does not detect empty array/or object

PR #669, opened 2019-10-18
Using local packages instead of global

oleg-andreyev:
It's possible that `mocha`, `webpack` and `xo` are missing in global NODE_PATH.
So need to use local package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,9 +25,9 @@
     "preversion": "npm test && git diff --exit-code --quiet",
     "postversion": "git push origin master && git push origin master --tags",
     "pretest": "npm run build",
-    "test": "mocha -r should",
-    "build": "webpack",
-    "posttest": "xo src lib bin"
+    "test": "node ./node_modules/mocha/bin/mocha -r should",
+    "build": "node ./node_modules/webpack/bin/webpack",
+    "posttest": "node ./node_modules/xo/cli.js src lib bin"
   },
   "dependencies": {
     "@babel/runtime": "^7.4.5",


willrowe:
Did you come across a specific issue with this?

Otherwise I believe this should be working fine since npm adds the local node_modules bin directory to the path before running scripts in the scripts section (which the dependencies should have their defined binaries copied to).

See the [npm docs for run-script](https://docs.npmjs.com/cli/run-script):
> In addition to the shell’s pre-existing PATH, npm run adds node_modules/.bin to the PATH provided to scripts. Any binaries provided by locally-installed dependencies can be used without the node_modules/.bin prefix.

oleg-andreyev:
I was executing scripts from PHPStorm was it was failing because it failed to find binaries.

willrowe:
Is it failing if you run the scripts directly from the command line?

package.json
```diff
@@ -25,9 +25,9 @@
     "preversion": "npm test && git diff --exit-code --quiet",
     "postversion": "git push origin master && git push origin master --tags",
     "pretest": "npm run build",
-    "test": "mocha -r should",
-    "build": "webpack",
-    "posttest": "xo src lib bin"
+    "test": "node ./node_modules/mocha/bin/mocha -r should",
+    "build": "node ./node_modules/webpack/bin/webpack",
+    "posttest": "node ./node_modules/xo/cli.js src lib bin"
```
RobLoach: Running scripts through npm will export node_modules/bin to PATH:
https://docs.npmjs.com/misc/scripts#path

oleg-andreyev:
I'll need to take a look at my setup on Vagrant machine where I use https://github.com/nvm-sh/nvm, because everything was working like a charm on MBP.

package.json
oleg-andreyev: Not longer an issue.

Issue #307, opened 2016-02-03
Expose filters so that they can be re-used / called from inside another?


sjuvonen:
Currently registered filters seem to be completely private and hidden from the exported Twig object, so they cannot be accessed at all outside templates. Usually template engines expose the filters, so I don't know why Twig.js is different in this manner. 

Is there a specific reason for not exposing filters or could it be done? It seems to be about only adding one line, Twig.exports.filters = Twig.filters, somewhere in the code.

plepe:
Sounds interesting, don't know if there's a reason to not expose it. Any thoughts by @justjohn or @evgenius about this?

PR #670, opened 2019-10-18
Export tests

oleg-andreyev:
Similar to #307

Use-case:
- I've found a bug in production e.g: #668
- I need to apply hotfix without forking.

I simply cannot do

```
// @see https://github.com/twigjs/twig.js/pull/668
window.Twig.filters.default = (value, _default = '') => {
    if (window.Twig.tests.empty(value)) {
        return _default;
    }

    return value;
};
```


RobLoach:
Seems useful, and having `tests` exported shouldn't negatively affect anything... Thanks @oleg-andreyev!

PR #674, opened 2019-10-29
Fix strict_variables issues

### LINKED ISSUE #629: strict_variables: true not working correctly
if I pass strict_variables: true to twig.js it detects this case correctly:
```
{{ test123.test }}
// Output: Can't access a key test on an null or undefined object.
```
but this case is not detected:
```
{{ test123 }}
// no error
```

in PHP this also throws: Variable "test123" does not exist.

btw: It would be great to have the file path as a part of the error message for all strict_variables errors.

toptalo:
Makes it to match twigfeddle errors output

Example of new errors:
- `Variable "foo" does not exist.`
- `Key "key" for object with keys "a, b, c" does not exist.`
- `Key "key" does not exist as the object is empty.`
- `Key "10" for array with keys "0, 1, 2" does not exist.`
- `Key "10" does not exist as the array is empty.`

Live example: https://twigfiddle.com/yfnyxt
*Tip: you need to switch main template to see different output*

Closes #629

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ var Twig = require("twig"),
 // This section is optional and used to configure twig.
 app.set("twig options", {
     allow_async: true, // Allow asynchronous compiling
-    strictVariables: false
+    strict_variables: false
 });
 
 app.get('/', function(req, res){
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ var Twig = require("twig"),
 // This section is optional and used to configure twig.
 app.set("twig options", {
     allow_async: true, // Allow asynchronous compiling
-    strict_variables: false
+    strictVariables: false
 });
 
 app.get('/', function(req, res){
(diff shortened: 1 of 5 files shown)


RobLoach:
Looking at the Twig PHP options, it uses `strict_variables`...

https://twig.symfony.com/doc/2.x/api.html

> strict_variables boolean

If set to false, Twig will silently ignore invalid variables (variables and or attributes/methods that do not exist) and replace them with a null value. When set to true, Twig throws an exception instead (default to false).

Unsure if it's best to go with the JavaScript option convention, or rolling with the one PHP uses.

toptalo:
I can revert this rename

toptalo:
I have an idea: what if we save `strict_variables` options param and rename all inner variable to camel case variant.

So here https://github.com/twigjs/twig.js/blob/master/src/twig.exports.js#L22
will be ` strictVariables: params.strict_variables || false,`

RobLoach:
That's a great idea... Would keep backward compatibility too :+1:

PR #675, opened 2019-10-31
Add support for 'deprecated' tag

josephineb:
Compiling the 'deprecated' tag will result in a warning message in the console.

Note: The 'deprecated' tag was added PHP Twig 2.6, see: https://twig.symfony.com/doc/2.x/tags/deprecated.html.

### DIFF
--- a/src/twig.logic.js
+++ b/src/twig.logic.js
@@ -41,7 +41,8 @@ module.exports = function (Twig) {
         embed: 'Twig.logic.type.embed',
         endembed: 'Twig.logic.type.endembed',
         with: 'Twig.logic.type.with',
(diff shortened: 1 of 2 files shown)


ericmorand:
+1 for using sinon in the test suite.

RobLoach:
This is great! Thanks so much :+1:

josephineb:
> This is great! Thanks so much

Thanks! Any idea when a new version will be released?

RobLoach:
Anything in particular you think is holding it back? The changelog looks pretty solid.

josephineb:
No, I don't think so.

PR #681, opened 2019-11-20
#468 Support for including arrays of templates

### LINKED ISSUE #468: {% include [] %}
Could we have a possibility to use include with an array of templates as it is the case for the php twig?

[http://twig.sensiolabs.org/doc/2.x/tags/include.html](http://twig.sensiolabs.org/doc/2.x/tags/include.html)

justafish:
Adds support for the array syntax used with include. See https://twig.symfony.com/doc/3.x/tags/include.html

Fixes #468

### DIFF
--- a/src/twig.logic.js
+++ b/src/twig.logic.js
@@ -794,8 +794,7 @@ module.exports = function (Twig) {
                         let files;
                         if (Array.isArray(file)) {
                             files = file;
-                        }
-                        else {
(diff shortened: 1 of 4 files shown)


RobLoach:
Thank you so much for adding this, @justafish. Been a feature that's been missed. Happy you're Twig.js-ing now! :wink: :rocket:

PR #687, opened 2019-12-09
fix autoescape for empty includes

tgabi333:
See test case, in the current release, it renders `[object Object]`

### DIFF
--- a/src/twig.core.js
+++ b/src/twig.core.js
@@ -709,7 +709,12 @@ module.exports = function (Twig) {
             return '';
         }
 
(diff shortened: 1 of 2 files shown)


src/twig.core.js
```diff
@@ -709,7 +709,12 @@ module.exports = function (Twig) {
             return '';
         }
 
-        return new Twig.Markup(escapedOutput.join(''), true);
+        const joinedOutput = escapedOutput.join('')
```
maxhelias: Just the semicolon for fix travis and it will be perfect :)

PR #690, opened 2019-12-11
Fix filters with empty string input

tbence94:
See the new tests. Filters should handle empty string inputs.

The `escape` and `nl2br` filters failed to do that. Current behavior:
```twig
""|escape   --> [object Object]
""|nl2br    --> throws error
```

I also added some tests for other filters, based on this commit:
https://github.com/twigjs/twig.js/commit/2021d34caea0fcb721674a2cf0e742d5b945fe5f

### DIFF
--- a/src/twig.filters.js
+++ b/src/twig.filters.js
@@ -405,7 +405,7 @@ module.exports = function (Twig) {
         },
 
         escape(value, params) {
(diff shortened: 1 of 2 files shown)


tgabi333:
:+1:  it would be super if we could have this and my fix in a new release @RobLoach please.

PR #702, opened 2019-12-22
raw filter on the empty string

drzraf:
The raw filter return a new Markup(content) which may itself return content

 if content is the empty string. The resulting output was `[Object]` instead of the expected empty string.

### DIFF
--- a/src/twig.core.js
+++ b/src/twig.core.js
@@ -1443,7 +1443,7 @@ module.exports = function (Twig) {
      */
 
     Twig.Markup = function (content, strategy) {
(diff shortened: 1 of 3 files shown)


serchavalos:
Thanks for taking care of this issue! I just noticed the regression too.

Do you have an estimated date for the next release?

serchavalos:
Hello @RobLoach! do you have an update yet?

RobLoach:
Thanks for the ping! Hope to get a release out today.

serchavalos:
Great! Thanks a lot

Issue #711, opened 2020-02-18
Simply tiwg.js in a browser


playinglife:
Hi,
I have spent many, many hours to try to make twig.js work in a browser. Unsuccessful :(
All I want is to be able to include the twig.js file and then render some templates.
I have many years of experience in development but it looks like this just got the better of me.

I have managed to make a twig.js version 0.9.8 to work but it has a bug in the "replace" filter.
I managed to make version 1.14.0 to load but when I'm rendering, the tags are not being replaced.

I include twig.min.js version 1.14.0 from here: https://asset-packagist.org/package/npm-asset/twig
The code I use:

`        var t="Are you {{ place }} at all?";
        Twig.twig({
            id: "posts",
            data: t,
        });
        alert(Twig.twig({ref: 'posts'}).render({'place': 'where'}));      
`

Thank you, I really appreciate any help,
Daniel

playinglife:
As a note: One needs to be careful when including a template especially when one also has TWIG templating on the server side.
If the template is not properly added into the code then the server side TWIG engine might process it too, besides the twig.js engine.
Ex: 
"Hello {{ name }} how are you."
If this would go into a view.html.twig template file then it should be:
{{' Hello {{ name }} how are you. '}}
so the php twig engine would not process it.

playinglife:
There might be a bug in the replace filter.

The code should display only the extension "JPG".
https://jsfiddle.net/playing_life/g96u3a4j/2/#&togetherjs=9v1FN3d6ap

RobLoach:
Does `(front):''` actually work?

playinglife:
Strangely the REPLACE works the first time but the second time looks like it still uses the data from the previous REPLACE call.

Ex: 
'Text x'|replace('x','1')         - this one calls the replace function with parameters "x" and "1"
'Text x'|replace('x','2')         - this one calls the replace function with parameters "x" and ("1" instead of "2". always the previous value)

Might be a variable scope issue.

willrowe:
This is due to a deeper issue with how expressions are parsed and tokenized. For some reason the  are being resolved to static values and cached instead of being kept dynamic and evaluated on each render.

- [twig.js example](https://codepen.io/willrowe/pen/mdxpYEg)

PR #712, opened 2020-02-18
Update twitter_backbone and add Browser example

RobLoach:
@playinglife This doesn't provide a multi-template sandbox, but could be a good place to demonatrate what you're trying to do...

Thoughts?

RE: #711

### DIFF
--- a/demos/browser/index.html
+++ b/demos/browser/index.html
@@ -1,0 +1,62 @@
+<!doctype html>
+<html lang="en-us">
+  <head>
+    <meta charset="utf-8">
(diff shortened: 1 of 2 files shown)


playinglife:
Thank you for your answer.

Basically it all started from this:
https://jsfiddle.net/playing_life/g96u3a4j/2/#&togetherjs=9v1FN3d6ap

The code should display only the extension "JPG". 
It looks like the REPLACE filter might have a bug.
And from here I just created all sorts of other issues that were related to other stuff.

PR #723, opened 2020-02-22
Simpler trimming logic

exeba:
Checking for a line composed only by whitespaces is redundant, the
trimmig would take care of it anyway.
It also makes the code harder to read and is probably slower in most
cases (http://jsben.ch/wT1mz).

### DIFF
--- a/src/twig.core.js
+++ b/src/twig.core.js
@@ -538,12 +538,9 @@ module.exports = function (Twig) {
                                 if (prevOutput.type === Twig.token.type.raw) {
                                     output.pop();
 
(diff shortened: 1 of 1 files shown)


RobLoach:
This may conflict with https://github.com/twigjs/twig.js/pull/722 ... Which solution do you propose we roll with?

exeba:
I split the 2 request because they are logically independent, but if you prefer I can create a new one with both of them.

#722 is definitely more important since it is a bugfix.

RobLoach:
Sounds good! Mind resolving the conflict? Then we can get it in. Thanks a lot!

PR #724, opened 2020-02-24
Make "js" escaped strings embeddable in JSON

dorian-marchal:
The ["js" escape strategy of Twig PHP](https://github.com/twigphp/Twig/blob/6384ba6c3ca7695ef0aec93bdf3b55c9bb18b5de/src/Extension/EscaperExtension.php#L247-L295) produces a JSON compatible escaping.
This is useful as it allows embedding JS escaped strings in JSON script.

For example, this can be used to templates JSON-LD scripts:

```html
<script type="application/ld+json">
{
  "@context": "http://schema.org/",
  "@type": "Person",
  "name": "{{ name }}",
}
</script>
```

Unfortunately, the twig.js implementation doesn't produce JSON compatible escaping.

For example:

```js
const { twig } = require('twig');
const compiledTemplate = twig({ autoescape: 'js', data: '{{ foo }}' });
compiledTemplate.render({ foo: '<foo>' });

// > \x3Cfoo\x3E
```

`\x3Cfoo\x3E` is not [valid JSON](https://tools.ietf.org/html/rfc7159#section-7) and can't be parsed by some JSON parsers.

With the same template/data, Twig PHP returns `\u003Cfoo\u003E`:

```php
<?php
$twig = new \Twig\Environment(new \Twig\Loader\ArrayLoader(), ['autoescape' => 'js']);
$twig->createTemplate('{{ foo }}')->render(['foo' => '<foo>']);

// > \u003Cfoo\u003E
```

In TwigPHP, The "js" escape strategy is defined there: https://github.com/twigphp/Twig/blob/6384ba6c3ca7695ef0aec93bdf3b55c9bb18b5de/src/Extension/EscaperExtension.php#L247-L295

Note: This PR replaces JSON-incompatible hexadecimal escape sequences (\x..) with JSON-compatible unicode escape sequences (\u00..).
It also prevents the escaping of some chars that are not escaped by TwigPHP.
This fixes specific use cases, but it doesn't make the escaping fully compatible with TwigPHP's.

I noticed that [`twing`](https://github.com/NightlyCommit/twing/blob/cb934c305bf68e8e507918275d477a1f7a45b387/src/lib/extension/core/filters/escape.ts#L52-L100) outputs the same string as TwigPHP.
Maybe you can help us, here, @ericmorand?

### DIFF
--- a/src/twig.filters.js
+++ b/src/twig.filters.js
@@ -431,10 +431,24 @@ module.exports = function (Twig) {
                     if (rawValue[i].match(/^[a-zA-Z0-9,._]$/)) {
                         result += rawValue[i];
                     } else {
+                        const char = rawValue.charAt(i);
                         const charCode = rawValue.charCodeAt(i);
 
-                        if (charCode < 0x80) {
-                            result += '\\x' + charCode.toString(16).toUpperCase();
+                        // A few characters have short escape sequences in JSON and JavaScript.
+                        // Escape sequences supported only by JavaScript, not JSON, are ommitted.
+                        // \" is also supported but omitted, because the resulting string is not HTML safe.
+                        const shortMap = {
+                            '\\': '\\\\',
+                            '/': '\\/',
+                            '\u0008': '\\b',
+                            '\u000C': '\\f',
+                            '\u000A': '\\n',
+                            '\u000D': '\\r',
+                            '\u0009': '\\t'
+                        };
+
+                        if (shortMap[char]) {
+                            result += shortMap[char];
                         } else {
                             result += Twig.lib.sprintf('\\u%04s', charCode.toString(16).toUpperCase());
                         }
--- a/test/test.core.js
+++ b/test/test.core.js
@@ -408,7 +408,7 @@ describe('Twig.js Core ->', function () {
                 data: '{{ value }}'
             }).render({
                 value: '<test>&</test>'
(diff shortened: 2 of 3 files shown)


ericmorand:
@dorian-marchal I can help yes. The `js` escaping strategy is implemented here in Twing:

https://github.com/NightlyCommit/twing/blob/cb934c305bf68e8e507918275d477a1f7a45b387/src/lib/extension/core/filters/escape.ts#L52

It has a dependency on `bin2hex` though, provided by Locutus PHP. I don't think that's a big issue.

dorian-marchal:
@ericmorand Thanks. :+1:

I see there is also a dependency on `utf8-binary-cutter`.

Both packages seem quite big:
- https://bundlephobia.com/result?p=utf8-binary-cutter@0.9.2
- https://bundlephobia.com/result?p=locutus@2.0.11

Twig.js probably doesn't work well for the edge cases these libs handle compared to JavaScript's native functions but maybe we could do without them. I believe Twig.js  doesn't try to be fully compatible with Twig PHP.

What do you think @RobLoach? Does this PR look OK to you as is?

dorian-marchal:
I rebased this PR on master. I'm available if you have questions about it @RobLoach. :+1:

RobLoach:
Thanks! Looks worth merging. If there's any follow ups, we can get them in a future PR. :+1:

PR #729, opened 2020-03-09
Add support for numbers to `starts with` and `ends with`

dorian-marchal:
TwigPHP's `starts with` and `ends with` also work for numbers.
This commit add support for them.

Also tests some more edge cases to make this operator a little more robust.


dorian-marchal:
My bad, this is not the way TwigPHP works at all. :smile:

Issue #737, opened 2020-04-20
Ternary operator overrides context


oleg-andreyev:
Test Case:
```
{% set classes = ['a', 'b'] %}
{% set classes = classes ? classes|merge(['c']) : '' %}
{{ dump(classes) }}
```

ER:
```
object(3) {
   [0]=> string(1) "a"
   [1]=> string(1) "b"
   [2]=> string(1) "c"
}
```

AR

```
number(3)
```

Issue caused by https://github.com/twigjs/twig.js/commit/6654458664f1197a5f3e843e322f5ba167ee1c54

It was fixed for `??` https://github.com/twigjs/twig.js/commit/7b3591dc9476b2207473eb1c43402a04751d8412

willrowe:
Fixed by #738.

PR #738, opened 2020-04-20
making sure that ternary expression does not change context

oleg-andreyev:
fixing #737 making sure that ternary expression does not change context + exporting functions for test purposes


oleg-andreyev:
@RobLoach please take a look

RobLoach:
🚀

Issue #741, opened 2020-06-02
Cannot load twig templates via Ajax from Firefox


shailpatels:
Latest versions of Firefox look like they need the XMLHttpRequest response type to be explicitly defined (in this case I think it should be set to "text"), otherwise, it'll think the response is XML and try to parse it which of course its not, twig.js is just pulling in text files to parse.

This causes a number of XML parsing errors whenever a Twig template get called with ajax.
![image](https://user-images.githubusercontent.com/12129065/83544862-8af7ca00-a4cc-11ea-8e6b-4bc48427a6c0.png)

Sample code 
```
Twig.twig({
    id: "test",
    href: "/example.twig",
    async: true
});

var out = Twig.twig({ ref: "test"}).render({ });
```

Does the render function also send XMLHttpRequests ? It looks it does which is done synchronously which is deprecated and cannot have an explicit response type.

Chrome appears to be fine 
Tested on Firefox 76.0.1 on MacOS & Firefox ESR Debian 10

shailpatels:
Closed via #742

PR #742, opened 2020-06-02
Specify MimeType to always use for ajax templates

MasterOdin:
For twig.js to be used in the browser, it requires the user to know that they either must name their templates as an ending that browsers recognized (e.g. `txt` or `html`) or to manually register a mime-type to the `.twig` extension. Otherwise, when fetching a template, a browser will attempt to parse it using `text/xml` which can cause issues depending on strictness (e.g. #741).

This PR changes it so that twig.js always specifies that templates received via XMLHttpRequest are `text/plain` and to be treated as such.

### DIFF
--- a/src/twig.loader.ajax.js
+++ b/src/twig.loader.ajax.js
@@ -40,6 +40,7 @@ module.exports = function (Twig) {
         };
 
         xmlhttp.open('GET', location, Boolean(params.async));
+        xmlhttp.overrideMimeType('text/plain');
         xmlhttp.send();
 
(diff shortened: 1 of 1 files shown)


RobLoach:
This makes sense to me. I can't think of issuing any other mime type through this. Thoughts Poly?

MasterOdin:
Any chance on seeing this get into a release in the near term? As is, without manually setting the content type through apache / nginx, twig.js is broken for the latest Firefox users for us.

PR #753, opened 2020-10-04
chore: use minified version of twig and update npmignore

jean-smaug:


RobLoach:
Was hoping to go the other way, and use the uncompressed version of the files! Just have the bundled one to ease use across other platforms!

willrowe:
Things will probably be moving towards converting things into modules and no longer providing bundled files in order to allow the most flexibility for using the bundler already being used in the applications that utilize this package.

PR #754, opened 2020-10-29
Add support for template arrays when using extends

justafish:
This currently doesn't work:

```
{% extends ['one.twig', 'two.twig'] %}
```

See https://twig.symfony.com/doc/2.x/tags/extends.html#dynamic-inheritance under "You can also provide a list of templates that are checked for existence"


RobLoach:
Thanks Sally! :rocket:

justafish:
@RobLoach you're welcome! Is there any chance of a release coming up soon? :grin:

PR #755, opened 2020-11-01
Update browser usage

obedparla:
After trying to use Twig on the browser with `twig({...})` and getting ` ReferenceError: twig is not defined` I realized I needed to do `Twig.twig({...})` instead. 

If this not always the case let me know so I can improve the documentation in this regard. 

It can be seen happening here: https://jsbench.me/6gkgwawsrs/1. Notice the import from unpkg.


obedparla:
That was a damn fast merge. Cheers!

PR #760, opened 2020-11-13
fix: fix lost context when calling a macro multiple times

### LINKED ISSUE #727: multiple macro calls leading to wrong output since v1.14.0
Hi,

when I define a macro in my twig template (incl. "_self" import) the first call of the macro will be fine. But starting with the second call the macro returns the value of the last parameter.

Example:

```
{% macro menuItem(text, class, url, isCurrent = false) %}
    <li><a href="#">dummy</a></li>
{% endmacro %}
{% import _self as menu %}

<nav>
    <ul class="menu">
        {{ menu.menuItem('Homepage', 'homepage', 'index.html', true) }}
        {{ menu.menuItem('Concept', 'concept', 'conzept.html', false) }}
        {{ menu.menuItem('About us', 'about-us', 'about-us.html') }}
    </ul>
</nav>
```

The first call will do a correct output of
`<li><a href="#">dummy</a></li>`

The second call will output `false` and the third call will output `about-us.html`.

_Note: before version 1.14.0 the output was correct._

mihkeleidast:
Should resolve #727.

### DIFF
--- a/src/twig.logic.js
+++ b/src/twig.logic.js
@@ -1006,6 +1006,7 @@ module.exports = function (Twig) {
                 state.macros[token.macroName] = function (...args) {
                     // Pass global context and other macros
                     const macroContext = {
+                        ...context,
                         _self: state.macros
                     };
                     // Save arguments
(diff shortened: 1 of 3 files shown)


RobLoach (approved):
Passing through the contexts seems like a good idea. Thanks!

mihkeleidast:
When can this be merged/released?

Issue #726, opened 2020-03-05
Date filter returns 1.1.1970 (unix 0) with UTC input again in version 1.15.0


VonSandberg:
Hi,
we just noticed that after we redeployed our app with bumped verison of twig npm package from version 1.14.0 to 1.15.0 some rendered dates started to be displayed as "1.1.1970".

For example: `{{ ORDER_DATE|date('j.n.Y') }}` where `ORDER_DATE = "2018-04-23T10:26:00.996Z"`
Note that Z (zulu timezone format) at the end, without it it works

I found this [old thread](https://github.com/twigjs/twig.js/issues/147) where is probably same issue addressed.

plepe:
I recently also noticed this regression (It was me you fixed the old issue, so I had a bit of a deja-vu). I looked into it, looks as this regression was added through a locutus update.

Date parsing uses the 'strtotime' function of locutus, and the current release (v2.0.14) has the same problem. This problem has been fixed though (locutusjs/locutus#432), but not been released yet.

For now I'll create a pull request with additional tests (checking the handling of time zones), and wait for the release of a new locutus version.

willrowe:
The bug in https://github.com/locutusjs/locutus was fixed and released in [v2.0.15](https://github.com/locutusjs/locutus/releases/tag/v2.0.15).

The dependency has been bumped up to v2.0.15 in #790 and should be released in v1.16.0.

PR #765, opened 2020-12-11
Fix date parsing with timezones

plepe:
This pull request adds tests for dates with timezone information, e.g.: `1995-12-17T08:24:00Z`, `1995-12-17T08:24:00+02:00`. The version with `Z` currently does not work. The problem is the strtodate function of locutus, which currently can't handle these strings (it has been fixed though, but not released yet).

Pull request which should eventually solve #726. Waiting for locutus release which incorporates locutusjs/locutus#432. I will update the pull request as soon as this release is available.

### DIFF
--- a/test/test.filters.js
+++ b/test/test.filters.js
@@ -301,6 +301,27 @@ describe('Twig.js Filters ->', function () {
             template.render().should.equal(stringDate(date));
         });
 
+        it('should recognize ISO 8601 date formats', function () {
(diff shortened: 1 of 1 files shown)


willrowe:
The bug in https://github.com/locutusjs/locutus was fixed and released in [v2.0.15](https://github.com/locutusjs/locutus/releases/tag/v2.0.15).

The dependency has been bumped up to v2.0.15 in #790 and should be released in v1.16.0.

These tests are useful to prevent regression, so I'm going to go ahead and merge them for release in v1.16.0 as well.

PR #770, opened 2021-01-13
Imported function PATHS.strip_slash() missing

mimidotsuser:
Exported function `strip_slash()` renamed to `stripSlash()` causing twigjs compilation command fail.

### DIFF
--- a/bin/twigjs
+++ b/bin/twigjs
@@ -24,7 +24,7 @@ while (args.length > 0) {
             return;
         case "--output":
         case "-o":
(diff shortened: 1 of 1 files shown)


stevefuentes:
The compile doesnt work without this fix, but it doesn't look like it's been merged into the build... Any ideas how to get this fix into the release?

mimidotsuser:
> The compile doesnt work without this fix, but it doesn't look like it's been merged into the build... Any ideas how to get this fix into the release?

I opted to switch to [PUG](https://github.com/pugjs/pug)

Issue #751, opened 2020-08-20
Variable gets undefined


wehmoen:
Hi,

I'm working on a pretty complex nodeJS using express and twig.js. Been using twig.js for over year now but encountered something strange today. 

In a template, lets call it `watch.twig` i extend `layout.twig`. In `watch.twig` I fill the blocks `meta`, `body` and `script`.  Using express I then run `res.render('watch`, {variable: 123})`

I use `{{ variable }}`  in all three blocks. The variable actually is written in the `meta` and `body` block. But not in the block called `script`.  When I used the `dump()` function in all three block i noticed that the variable is not availbale in this specific block. I upgraded from 0.10.3 to 1.10.4 today and that seems to be the only issue. Before I upgrade twigjs any further I wanted to ask if anyone here ever came across something like this. 

Kind regards

-- The actual templates are much more complex than what I describe here. Just for simplicity:

`layout.twig`

```twig

<div>{% block meta %}{% endblock %}</div>
<div>{% block body %}{% endblock %}</div>
<div>{% block script %}{% endblock %}</div>

```

`watch.twig`

```twig

{% extends './layout.twig' %}

{% block meta %}{{ variable }}{% endblock %}
{% block body %}{{ variable }}{% endblock %}
{% block script %}{{ variable }}{% endblock %}

```

Expected Output:

```
<div>123</div>
<div>123</div>
<div>123</div>
``` 

Actual Output:

```
<div>123</div>
<div>123</div>
<div></div>
```

Simplified express route: 

```js
router.get('/path', (req, res) => {

   res.render('watch', {variable: 123});

});
```

Versions: 

TwigJS: 1.10.4
Express: 4.17.1

wehmoen:
Been able to track it down to one line in the `layout.twig` that uses a macro. A simple macro, that based on the date of the year returned a different string caused these issues. When I removed the macro everything works as expected.  

```twig
{% macro getLogo() %}
  {% set logo_normal = '/img/logo.png' %}
  {% set logo_holiday = '/img/logo_holiday.png' %}
  {% set logo_sylvester = '/img/logo_sylvester.png' %}
  {% set today = ''|date("d") %}
  {% set month = ''|date("m") %}

  {% if month == '12' %}
    {% if today == "24" or today == "25" or today == "26" %}
      {{ logo_holiday }}
    {% elseif today == '31' %}
      {{ logo_sylvester }}
    {% else %}
      {{ logo_normal }}
    {% endif %}
  {% else %}
[...]

Issue #692, opened 2019-12-12
Macro mutates data object in v1.14.0


kachurun:
I render templates in browser from strings like this:

```
const imp = '{% import _self as _macro %}';
const macro = `
	{% macro testMacro(global) %}
		<span>testMacro content</span>
	{% endmacro %}
`;
const template = `
	<div>
		{{ _macro.testMacro(_context) }}

		global.example: {{ global.example ?? 'null' }}
		example: {{ example ?? 'null' }}
                dump: {{ dump(_context) }}
	</div>
`;

const data = {
    example: 'test'
};

Twig.cache();

const tw = Twig.twig({
    id: 'test.html', 
    data: `${ macro }${ imp } ${ template }`
});

tw.renderAsync(data).then((output) => {
    console.log('output', output)
});
```

And its broken with version 1.14.0. All variables puts in `global` object ('global' is name of first argument in macro).

I get 
```
        <div>
            <span>testMacro content</span>
            global.example: test
            example: null
            dump: 
            object(2) {
                [_self] => object(1) {
                    [testMacro] => function()
                }
              
                [global]=> object(2) {
                    [example] => string(4) "test"
                    [_macro] => object(1) {
                        [testMacro] => function()
                    }
                }
            }
        </div>
```

With version 1.13.3 all works fine.

willrowe:
- [TwigPHP example](https://twigfiddle.com/rgaibl)
- [twig.js example](https://codepen.io/willrowe/pen/oNqoOWp)

PR #773, opened 2021-02-14
fix macro changing context in loop

mihkeleidast:
Possibly related: #692, #751.


mihkeleidast:
@RobLoach any chance this'll get a review soon-ish? Macros in (nested) loops do all kinds of crazy things right now & we would very much appreciate if this would get merged & released.

FelipeMarcos:
@RobLoach please merge? 🙏

Calling macros inside other macros is broken right now because parameters are being mutated to undefined.

PR #775, opened 2021-02-26
add some necessary exceptions to strict_variables = true handling

### LINKED ISSUE #679: Regression: Enabled strict_variables throw unnecessary errors
The changes from #629 seem to throw unnecessary errors.
Following template should always run even if `lie` isn't defined and `strict_variables` is enabled - just the output should change:
``The {{ baked_good }} is a{% if lie is defined and lie %} lie {% else %} reality {% endif %}!`` 

However, with `strict_variables`  enabled it fails with `TwigException: Variable "lie" does not exist.`
Running Example: https://jsfiddle.net/jsz9uy3h/6/

If you use https://cdn.jsdelivr.net/npm/twig@1.13.3/twig.min.js in above fiddle it works.

For the moment I can only think of introducing context awareness into the `parse` function of `Twig.expression.type.variable`. However this seems like a good way to clutter things :|

d--j:
fixes #679

This solution is based on https://github.com/twigjs/twig.js/issues/679#issuecomment-675341074 but handles more cases.

### DIFF
--- a/src/twig.expression.js
+++ b/src/twig.expression.js
@@ -847,7 +847,25 @@ module.exports = function (Twig) {
                 return Twig.expression.resolveAsync.call(state, context[token.value], context)
                     .then(value => {
                         if (state.template.options.strictVariables && value === undefined) {
-                            throw new Twig.Error('Variable "' + token.value + '" does not exist.');
+                            let skipException = false;
+                            if (token.peek) {
+                                const {peek} = token;
+                                if (peek.type === Twig.expression.type.filter && peek.value === 'default') {
+                                    skipException = true;
+                                }
+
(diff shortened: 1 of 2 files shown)


antoineveldhoven (approved):

antoineveldhoven (approved):

willrowe:
@antoineveldhoven please do not spam PRs with reviews. If you have something to contribute, elaborate with a comment.

willrowe (approved):

willrowe:
I believe that the correct way to handle this would be to do the strict variables check when the variable is being used and not when parsing it, but that is a far more complex undertaking. I think this is good enough to address the current issue for now.

PR #780, opened 2021-05-05
feat: context in filters

simoneb:
Provides context to custom filters as third parameter:

```js
function filter(input, params, context) {
 // ...
}
```

This is often useful when implementing custom filters whose logic may need access to the rendering context. An easy workaround is to provide the context (or part of it) as a filter parameter, but this option doesn't require any syntax changes in the templates.

### DIFF
--- a/test/test.extends.js
+++ b/test/test.extends.js
@@ -1,156 +1,173 @@
+const sinon = require('sinon')
 const Twig = require('../twig').factory();
 
 const {twig} = Twig;
 
 describe('Twig.js Extensions ->', function () {
-    it('should be able to extend a meta-type tag', function () {
-        const flags = {};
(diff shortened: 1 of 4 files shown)


test/test.extends.js
```diff
@@ -1,156 +1,173 @@
+const sinon = require('sinon')
 const Twig = require('../twig').factory();
 
 const {twig} = Twig;
 
 describe('Twig.js Extensions ->', function () {
-    it('should be able to extend a meta-type tag', function () {
```
simoneb, line 6: the change in this file is simply:

- nesting the existing tests under a describe blog specific for extending tags
- add test for custom filter (doesn't look like they were tested before, except in async tests)

Issue #794, opened 2021-09-01
Cache false not working for inline includes


ostrojs:
When a template is used in a page, the modifications are not reflected.

christian-beckmann:
This is true. Setting 

// by default
Twig.cache = true;
to 
Twig.cache = false;

renders includes (and embeds) correclty.

christian-beckmann:
Despite passing {cache: false}, the cache is used.

willrowe:
@ostrojs can you provide an example [using this template](https://codepen.io/pen?template=LYQawea) at a starting point?

PR #795, opened 2021-09-01
fixes for issue #794

ostrojs:
If need to save a template, make sure cache is turned on.


willrowe:
This doesn't really explain the issue fully and has no tests. Please comment further on #794 so that the bug can be confirmed.

PR #800, opened 2021-11-08
Refactor/change to classes

dmytro-lt:


willrowe:
Thank you for putting this together. While things will be moved towards using more classes, unfortunately, this contains too many changes to be able to be merged.

PR #817, opened 2022-06-22
Add GitHub actions test workflow

### LINKED ISSUE #815: Add CI with GitHub Actions
This would run the test suite.

willrowe:
This adds a workflow to run the test suite. It currently only runs on pull requests, but could be expanded if needed.

This resolves #815.


RobLoach (approved):
Thanks so much for all the movement here, @willrowe!

PR #823, opened 2022-08-31
Add namespace support to `source` function

### LINKED ISSUE #442: Namespaces with source() do not resolve

willrowe:
This adds support for paths with namespaces to the `source` function. Fixes #442.


RobLoach (approved):
Thanks @willrowe! Looks great :+1:

PR #784, opened 2021-05-12
Fix #708 - Allow multiple includes of a template that embeds another …

sdussaut:
Multiple template includes that embed other template breaks the compilation.

This PR is basically @JamesB797 's fix.

Issue: https://github.com/twigjs/twig.js/issues/708

PR #828, opened 2022-09-21
Fix multiple includes with embeds

### LINKED ISSUE #708: Crashes on including a template twice that embeds another template
I have some complex twig templating from a php twig project that crashes in twig js. Here is the minimum code required to trigger the crash.

```js
const Twig = require('twig');
const twig = Twig.twig;

twig({
  id: 'poc2',
  data: ``
});

twig({
  id: 'poc1',
  data: `
    {% embed 'poc2' %}{% endembed %}
  `
});

const poc = twig({
  data: `
    {% include 'poc1' %}
    {% include 'poc1' %}
  `
});

console.log(poc.render());
```

I'm going to try to fix it and submit a PR, but if someone knows a solution or one is already in the works please let me know.

Thanks!

willrowe:
This PR supersedes #784. After creating a failing test and debugging the issue, it turns out that the `id` should have never been set on the temporary template instance, since that stored it in the registry and allowed subsequent calls to the loader to return the temporary instance instead of the correct instance set before the embed.

Fixes #708

### DIFF
--- a/src/twig.logic.js
+++ b/src/twig.logic.js
@@ -1239,7 +1239,6 @@ module.exports = function (Twig) {
                     .then(fileName => {
                         const embedOverrideTemplate = new Twig.Template({
                             data: token.output,
-                            id: state.template.id,
                             base: state.template.base,
                             path: state.template.path,
                             url: state.template.url,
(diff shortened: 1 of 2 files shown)


RobLoach:
Thanks so much, Will :rocket:

PR #829, opened 2022-09-22
Use src/twig.js as package's main script instead of the compiled twig.js

### LINKED ISSUE #825: Use src/twig.js instead of built twig.js
Packaging it up as twig.js and having Node.js use the compiled twig.js makes debugging extremely difficult. I'm proposing shipping the src directory with the package, and having "main" reference src/twig.js instead.

RobLoach:
Fixes #825

### DIFF
--- a/lib/compile.js
+++ b/lib/compile.js
@@ -1,7 +1,7 @@
 const FS = require('fs');
 const minimatch = require('minimatch');
 const WALK = require('walk');
-const Twig = require('../twig');
+const Twig = require('..');
 const PATHS = require('./paths');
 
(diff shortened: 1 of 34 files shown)


willrowe:
While this make debugging application code easier, I guess I was thinking in #825 that it would make debugging the tests easier as well. Can you update the tests use the source code instead of the compiled `twig.js` file too?

RobLoach:
Good call :+1:

willrowe:
Looks great, thanks @RobLoach!

willrowe (approved):

PR #839, opened 2022-10-05
Add support for `with` tag without context or `only` keyword

### LINKED ISSUE #774: Tag "with" not working: TwigException: Unable to parse 'with'
Tag `with` not working at all.
```
{% with %}
  {% set foo='bar' %}
   This is foo: {{foo}}
{% endwith %}
``` 
And got error:
```
TwigException: Unable to parse 'with'
```

willrowe:
This resolves #774.


willrowe:
@RobLoach can you review this so we can get it merged?

RobLoach (approved):

PR #846, opened 2022-12-05
Use v3 of `actions/checkout`

willrowe:
This fixes the deprecation notice in the actions by using NodeJS v16 instead of v12.

@RobLoach I'm going ahead and merging this since it's so minor.


RobLoach:
Wooo!

PR #847, opened 2022-12-05
Test on more node versions

willrowe:
This supersedes #835 and tests up to Node v18.

I ended up creating a proxy script to run webpack and detect when Node v17+ is being used and add the Node flag accordingly.


RobLoach:
Nice work!

PR #850, opened 2023-01-03
Fix passing context around

### LINKED ISSUE #840: Bug with context scoping
While working on #839, I discovered a bug in how the `ParseState` context is being passed around that is causing the `with` tag to not work as intended. Unfortunately, the tests do not currently test whether the context within the tags is isolated and does not affect the context outside of them.

This is most likely causing other issues with context not being properly scoped.

I am working on a PR to simplify how the context is handled and make scoping more explicit.

willrowe:
This fixes #840 by allowing the context to be passed when initially creating an instance of `ParseState`. The old way of passing the context when calling `parse` is still supported to prevent BC breaks, but it no longer used internally.

### DIFF
--- a/src/twig.core.js
+++ b/src/twig.core.js
@@ -1325,9 +1325,9 @@ module.exports = function (Twig) {
         params = params || {};
 
         return Twig.async.potentiallyAsync(template, allowAsync, () => {
-            const state = new Twig.ParseState(template, params.blocks);
+            const state = new Twig.ParseState(template, params.blocks, context);
 
-            return state.parseAsync(template.tokens, context)
+            return state.parseAsync(template.tokens)
                 .then(output => {
                     let parentTemplate;
                     let url;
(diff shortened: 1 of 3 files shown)


fernakr:
Hi there, sorry if this is not a good place to comment this but was curious if you had a ballpark estimate on when 1.16 will be released. Mostly just need to know if I can wait it out or pivot to a different approach on the issue I'm running into currently.

Thanks for your hard work!

willrowe:
@fernakr is there a specific issue that has already been fixed, but not released?

This is the final PR that needs to be merged before v1.16.0 can be released, I'm just waiting on @RobLoach to review at this point.

fernakr:
Gotcha - it was the nested embeds problem: https://github.com/twigjs/twig.js/issues/708

willrowe:
@RobLoach if you could take a look at this and then just go ahead and merge and publish v1.16.0 that would be great. It would be great to go ahead and get it released since it has a lot of long standing bugs fixed in it.

RobLoach:
@willrowe Thanks for the ping! This is looking really good, and will do some rigorous testing, along with getting the release out later on today :rocket:

fernakr:
Awesome thanks to you both!

RobLoach:
It was all @willrowe!!!! MVP of this release for sure :rocket:

PR #859, opened 2023-03-11
Adds the (string) position of tokens to the token tree.

synga-nl:
This PR adds the position to the tokens in the token tree. 

Example found in the following string:
`I want to {{ try }} a more {% if advanced | length > 3 %}{{ variable }}{% endif %} template {% set unit = 2 %}{# This is a comment #}{{ variable_after_comment }}`

```
  {
    type: 'raw',
    value: ' template ',
    position: { start: 82, end: 92 }
  },
```

This means the raw text ` template ` can be found between position 82 and 92 in the given template.

This is one of a series of PRs I want to make to parse through tokens and to act on these tokens. Like the NodeTraverser of Twig in PHP.

### DIFF
--- a/src/twig.core.js
+++ b/src/twig.core.js
@@ -325,9 +325,8 @@ module.exports = function (Twig) {
      */
     Twig.tokenize = function (template) {
         const tokens = [];
-        // An offset for reporting errors locations in the template.
-        let errorOffset = 0;
-
+        // An offset for reporting errors locations and the position of the nodes in the template.
+        let currentPosition = 0;
         // The start and type of the first token found in the template.
         let foundToken = null;
         // The end position of the matched token.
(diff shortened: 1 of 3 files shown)


willrowe:
This looks pretty good. I don't see any reason why it can't be merged, any thoughts @RobLoach?

willrowe (approved):

synga-nl:
@willrowe or @RobLoach is it possible to create a minor version of this on NPM? That makes testing the package in other applications way more easy for the other PRs (token traverser).

willrowe:
I was going to add this to the v1.17.0 milestone that I will be choosing fixes/features for in the next couple of weeks, so I'm not sure if we'd do a release for just this or not.

PR #870, opened 2023-09-25
Allow multiple spaces after elseif statement.

### LINKED ISSUE #688: elseif with two spaces triggers exception
The following code triggers a parse error in twig.js (this compiles successfully in its PHP counterpart). Note I've only tested this for the `elseif` tag. It could affect other tags as well. Removing the extra space after `elseif` fixes the parse error.
```
Hi there
{% if true %}
Yup
{% else %}
{% if 1 > 2 %}
This should not print
{% elseif  5 > 2 %}
  This should print
{% endif %}
{% endif %}
```
**Expected behavior:** code compiles successfully despite the extra space after `elseif`
**Actual behavior:** code returns the following error:
```
TwigException: Unable to parse 'elseif  5 > 2'
```
Node version 12.13.1
Twig version 1.12.2

antoineveldhoven:
Fixes #688

### DIFF
--- a/src/twig.logic.js
+++ b/src/twig.logic.js
@@ -122,7 +122,7 @@ module.exports = function (Twig) {
              *  Format: {% elseif expression %}
              */
             type: Twig.logic.type.elseif,
-            regex: /^elseif\s?([^\s].*)$/,
+            regex: /^elseif\s*([^\s].*)$/,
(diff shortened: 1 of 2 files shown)


willrowe:
@antoineveldhoven thank you!

PR #874, opened 2023-10-11
Arrow functions

### LINKED ISSUE #652: Add filter, map and reduce features

antoineveldhoven:
Aiming to resolve https://github.com/twigjs/twig.js/issues/652

I'm not (yet) quite convinced if this is the way to go; would really like to get some discussion going on how to approach this and eventually get this fixed and merged.

Implementation of arrow functions in filters:

    sort
    filter
    map
    reduce

Object manipulation is not really supported; this implementation accepts a arrowFunction (Twig syntax) which is compiled and parsed as a new (runtime) template. The return values (e.g. output) of the arrowFunction are delegated to their respective JS functions.

As previously stated: this is merely a functional (read dirty) workaround; but for most of the use cases it will do the trick.

### DIFF
--- a/src/twig.expression.js
+++ b/src/twig.expression.js
@@ -35,6 +35,7 @@ module.exports = function (Twig) {
             unary: 'Twig.expression.type.operator.unary',
             binary: 'Twig.expression.type.operator.binary'
         },
+        arrowFunction: 'Twig.expression.type.arrowFunction',
         string: 'Twig.expression.type.string',
         bool: 'Twig.expression.type.bool',
         slice: 'Twig.expression.type.slice',
@@ -71,6 +72,7 @@ module.exports = function (Twig) {
         // What can follow an expression (in general)
         operations: [
             Twig.expression.type.filter,
+            Twig.expression.type.arrowFunction,
             Twig.expression.type.operator.unary,
             Twig.expression.type.operator.binary,
             Twig.expression.type.array.end,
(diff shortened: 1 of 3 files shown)


willrowe:
I'm in the process of reviewing this, but I'm not sure if this approach is going to work. I think the body of the arrow function will need to be parsed as an expression. I'm trying to come up with a more general way for it to be compiled so a simple function can be called by the filters that support it.

ericmorand:
I'm not sure this will work.

It is perfectly possible to register filters (and I assume functions and tests) that return promises:

```js
import Twig from "twig";

import pkg from 'twig';
const {twig} = pkg;

Twig.extendFilter('asyncUpper', (value) => Promise.resolve(value.toUpperCase()));

const t = twig({
    data: '{{ "a"|asyncUpper}}'
});

t.renderAsync().then(console.log); // outputs "A"
```

How would the arrow function behave in this case?

```twig
{{ [4, 5]|sort((a, b) => a|asyncUpper< b ? -1 : 1)|join() }}
```

I think some tests need to be added to ensure that the arrow function support expression and async expressions specifically. The current tests suite provided by this PR is far from covering all the possible expressions that Twig accepts as the body of the arrow function. Moreso because Twig.js supports asynchronous expressions.

michaeldietiker:
Any updates on arrow function support?

ImKubass:
Any updates on this?

Manfredik:
When ??

willrowe:
@omarius it does not function correctly as-is unfortunately, so I do not think it will be merged any time soon.

antoineveldhoven:
Closing this PR.
PR #957 is a better solution (still needs to be confirmed).

PR #879, opened 2023-10-22
Allow colon inside Twig.expression.type.key.brackets.

### LINKED ISSUE #854: accessing an array element with a colon
I'm passing in a hydra json response, which includes an element with a key of 'hydra:member'.

My twig template is

```twig
<h3>
    {{ data['@type'] }}
</h3>
{% for d in data['hydra:member'] %}
{{ d.id }}
{% endfor %}
```

This throws an error

TwigException: Twig.expression.type.array.start cannot follow a Twig.expression.type.variable at template:4 near '[...'
 ERROR  Failed to compile with 1 errors                                                                                                                                                       Syntax Error: Unexpected token u in JSON at position 0
    at JSON.parse (<anonymous>)

My solution is to set a temporary variable and use it, but I'm wondering if there's a better way, or if this can be fixed.

```twig
{% set hack = 'hydra:member' %}
{% for d in data[hack] %}
{{ d.id }}
{% endfor %}
```

Thanks.  I think this library rocks.

antoineveldhoven:
Fixes #854

Not sure why the colon is needed in de regexp for Twig.expression.type.key.brackets, @willrowe can you elaborate?

If it is not needed, removing the colon in the regexp will fix this issue.

### DIFF
--- a/src/twig.expression.js
+++ b/src/twig.expression.js
@@ -913,7 +913,7 @@ module.exports = function (Twig) {
         },
         {
             type: Twig.expression.type.key.brackets,
-            regex: /^\[([^\]:]*)\]/,
+            regex: /^\[([^\]]*)\]/,
             next: Twig.expression.set.operationsExtended.concat([
                 Twig.expression.type.parameter.start
             ]),
--- a/test/test.expressions.js
+++ b/test/test.expressions.js
@@ -325,6 +325,11 @@ describe('Twig.js Expressions ->', function () {
             const testTemplate = twig({data: '{% if a is defined and a %}true{% else %}false{% endif %}'});
             testTemplate.render({a: ['value']}).should.equal('true');
         });
+
+        it('should be able to access array elements with colons', function () {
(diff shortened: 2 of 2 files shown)


willrowe:
It appears that came for #362, but the tests from that PR appear to be passing, so it may have not been necessary to alter the regex originally. Can you double-check that PR and ensure that slicing still works?

antoineveldhoven:
> It appears that came for #362, but the tests from that PR appear to be passing, so it may have not been necessary to alter the regex originally. Can you double-check that PR and ensure that slicing still works?

Thanks for elaborating @willrowe !

As far as I can tell, the regexp with the colon is defined in Twig.expressions.js (line 578) Twig.expression.type.slice, so it is part of type slice and type.key.brackets are not used in this expression.

LGTM!

RobLoach (approved):
The tests are passing. Think this is good to go, @willrowe ?

willrowe:
It should be, I think @antoineveldhoven is right and the regex must have been updated by mistake in the previous PR>

PR #881, opened 2023-10-24
Support variables in slice filter shorthand

### LINKED ISSUE #769: The slice filter shorthand doesn't support variables
The slice filter type regex is `regex: /^\[(\d*:\d*)\]/,` which only matches something like `[10:]` but not `[length:]`. I can just use `|slice(length)` and `|slice(0, length)` but it's a little unexpected coming from php twig.

antoineveldhoven:
Resolves #769

### DIFF
--- a/src/twig.expression.js
+++ b/src/twig.expression.js
@@ -576,14 +576,14 @@ module.exports = function (Twig) {
         },
         {
             type: Twig.expression.type.slice,
-            regex: /^\[(\d*:\d*)\]/,
+            regex: /^\[(-?\w*:-?\w*)\]/,
             next: Twig.expression.set.operationsExtended,
             compile(token, stack, output) {
                 const sliceRange = token.match[1].split(':');
 
                 // SliceStart can be undefined when we pass parameters to the slice filter later
-                const sliceStart = (sliceRange[0]) ? parseInt(sliceRange[0], 10) : undefined;
-                const sliceEnd = (sliceRange[1]) ? parseInt(sliceRange[1], 10) : undefined;
+                const sliceStart = sliceRange[0];
+                const sliceEnd = sliceRange[1];
 
                 token.value = 'slice';
                 token.params = [sliceStart, sliceEnd];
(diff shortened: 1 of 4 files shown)


RobLoach (approved):
Oh this is nice. Thanks a lot. Looks good to me. Anything else we should address before merging?

antoineveldhoven:
> Oh this is nice. Thanks a lot. Looks good to me. Anything else we should address before merging?

Thanks for the review, this commit should merge perfectly as-is.
Could you or @willrowe take a look at #874, we really need this to me merged for our project(s).

src/twig.expression.js
```diff
@@ -576,14 +576,14 @@ module.exports = function (Twig) {
         },
         {
             type: Twig.expression.type.slice,
-            regex: /^\[(\d*:\d*)\]/,
+            regex: /^\[(-?\w*:-?\w*)\]/,
```
willrowe, line 579: It looks like you're adding support for negative values here, but there are no tests for this. Please add tests.

src/twig.expression.js
```diff
@@ -596,11 +596,39 @@ module.exports = function (Twig) {
...
+            parse(token, stack, context) {
                 const input = stack.pop();
-                const {params} = token;
+                let {params} = token;
                 const state = this;
 
+                if (parseInt(params[0], 10).toString() === params[0]) {
+                    params[0] = parseInt(params[0], 10);
+                } else {
+                    const value = context[params[0]];
```
willrowe, line 607: This only allows very simple variable names. Technically PHP Twig supports full expressions as slice parameters, like accessing a member value of a map or a math expression. However, I'm just mention this to point it out. I'm fine with allowing this simple case to push things forward and don't thing it should be a blocker to get this merged.

test/test.expressions.js
```diff
@@ -455,6 +455,12 @@ describe('Twig.js Expressions ->', function () {
             output.should.equal('23');
         });
 
+        it('should support slice shorthand (full form) with variables as arguments', function () {
```
willrowe, line 458: Please add more test cases to cover both the optional first and second parameters.

test/test.filters.js
```diff
@@ -640,6 +640,10 @@ describe('Twig.js Filters ->', function () {
             const testTemplate = twig({data: '{{ \'12345\'|slice(1, 2) }}'});
             testTemplate.render().should.equal('23');
         });
```
willrowe, line 642: Please add more test cases to cover both the optional first and second parameters.

willrowe (changes requested):

src/twig.expression.js
antoineveldhoven: Thanks for pointing this out!
I wasn't aware of it.

For now I think it is OK to allow these simple cases; I'll create an issue in order to support full expressions; but have to look into the current implementation in PHP before doing that.

antoineveldhoven:
@willrowe Thanks for the extensive review; I will add the requested tests as soon as possible.

antoineveldhoven:
@willrowe Can you take a look at my revision of this PR?
I;ve added an additional check for negative length parameter in the slice filter as in [PHP: array_slice](https://www.php.net/manual/en/function.array-slice.php )

willrowe (approved):

JamesB797:
Nice, thank you for addressing this!

willrowe:
@RobLoach can you go ahead and release v1.17.0? Everything in the milestone has been completed, so we might as well release instead of adding more at this point.

PR #883, opened 2023-11-13
Array comparisons

### LINKED ISSUE #691: `in` operator doesn't work like Twig PHP when comparing arrays
For example, this evaluates to "true" in PHP but "false" in twig.js:

```twig
{{ ["foo", "bar"] in [['foo', 'bar'], ['qux', 'baz']] ? 'true' : 'false' }}
```

### LINKED ISSUE #882: [BUG] Comparison of Array with boolean
Comparison of Array with boolean will only be true if Array.length == 1.
For arrays with more than one element this will render incorrect.

[TwigPHP example](https://twigfiddle.com/ppn4ac)
[twig.js example](https://codepen.io/antoineveldhoven-the-typescripter/pen/ExrvNvB)

antoineveldhoven:
Attempting to fix #882 and fix #691 

Please note that comparisons are not strict.
Maybe we should add an equals prototype for Arrays like in [this Stack Overflow post](https://stackoverflow.com/questions/7837456/how-to-compare-arrays-in-javascript#14853974)

### DIFF
--- a/src/twig.expression.operator.js
+++ b/src/twig.expression.operator.js
@@ -178,12 +178,13 @@ module.exports = function (Twig) {
         }
 
         if (operator !== 'in' && operator !== 'not in' && operator !== '??') {
-            if (a && Array.isArray(a)) {
-                a = a.length;
-            }
-
-            if (operator !== '?' && (b && Array.isArray(b))) {
-                b = b.length;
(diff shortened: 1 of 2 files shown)


willrowe:
I'm going to close this because it's a couple of years old at this point and it is trying to fix two different issues. If someone wants to open a PR with more of a focused fix, it should be easier to review separately.

PR #884, opened 2023-11-15
Add trans filter to replicate Symfony/Drupal trans

dan2k3k4:
We have a custom Drupal theme that uses Storybook.
When we load up Storybook, the trans filter in the twig PHP files throw errors.
Thus, we add a simple `trans` filter to twig.js to get around this issue.

### DIFF
--- a/src/twig.logic.js
+++ b/src/twig.logic.js
@@ -43,7 +43,8 @@ module.exports = function (Twig) {
         endembed: 'Twig.logic.type.endembed',
         with: 'Twig.logic.type.with',
         endwith: 'Twig.logic.type.endwith',
-        deprecated: 'Twig.logic.type.deprecated'
+        deprecated: 'Twig.logic.type.deprecated',
+        trans: 'Twig.logic.type.trans'
     };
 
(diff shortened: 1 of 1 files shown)


ericmorand:
The `trans` filter is not part of the core Twig specification. I think that it should not be put in twig.js code base, but should be provided by a third-party library.

dan2k3k4:
@ericmorand do you have a suggestion on how that should be implemented?

We've been using a patched version of `twig.js` for a few years with the change from this PR, then we `npm run build` and `npx patch-package twig` to get our patch file for the current version of twig.js that we're using.

For our needs, it has been working fine, I figured I would just provide the solution in case others want to do something similar.

ericmorand:
twig.js supports registering custom tags: 

https://github.com/twigjs/twig.js/wiki/Extending-twig.js-With-Custom-Tags

Does it fit your need?

RobLoach:
@dan2k3k4 There is also this: https://github.com/JohnAlbin/drupal-twig-extensions

willrowe:
Closing this since `trans` is not part of core Twig.

PR #886, opened 2023-11-16
Fix package-lock.json for compatibility with Node.js 10+

RobLoach:
Needs to be compatible with older Node.js's?


RobLoach:
@willrowe WAT?

willrowe:
@RobLoach what happened? I don't think anything changed in the dependencies so the lock file shouldn't have changed besides the version number?

RobLoach:
I believe what happened is that when I ran the tests locally with Node.js 20, it updated the lock file... Which may or may not work on Node.js 10 :upside_down_face: .... I'll issue a point fix with this one.

willrowe:
Oh my, didn't even realize that could happen. Thanks for the release!

PR #907, opened 2025-04-08
Security updates

willrowe:
This updates dependencies that have reported vulnerabilities. The only dependency that could not be updated is `xo`, since starting in v0.40.0 it only works with ESM projects.

This also drops support for Node v10 and v11.


willrowe:
The support matrix for Node with dependencies is unfortunately too complex to update at this time.

PR #913, opened 2025-05-28
Provide PHP style output of booleans by specifying option.

### LINKED ISSUE #88: "Twig supports binary operations (+, -, *, /, ~, %, and, or)" incorrect print of boolean
Twig.php follows the same rules for printing booleans as PHP does.

In PHP "false" is printed as empty string. "true" is printed as 1

```
{{ 1 and 0 }}
```

Print:

```
0
```

should print:

```

```

I'm not sure how to fix this. 
I should change Twig.expression.operator.parse to add the PHP-like strings, but isn't that function both used for outputting text and the evaluating operator expressions?

I could also, make Twig.expression.operator.parse add the proper types to the stack and then parse the output at the end (the expression "a && b" add a 1 if true and 0 if false), but I don't understand the code well enough to see if this is a good idea or not.

antoineveldhoven:
I'm not sure if this is the way to go, since this is a very old issue.
Please share your thoughts @willrowe && @RobLoach about this fix.

Aiming to fix #88 

Since this change breaks every boolean test and / or comparison, I've introduced an option in order to enable PHP style boolean outputs.

### DIFF
--- a/src/twig.core.js
+++ b/src/twig.core.js
@@ -695,6 +695,18 @@ module.exports = function (Twig) {
      */
     Twig.output = function (output) {
         const {autoescape} = this.options;
+        const {phpStyleBooleans} = this.options;
+
+        // Conform Javascript boolean to PHP boolean
+        if (phpStyleBooleans) {
+            output = output.map(str => {
+                if (typeof str === 'boolean') {
+                    str = str ? '1' : '';
+                }
+
+                return str;
+            });
+        }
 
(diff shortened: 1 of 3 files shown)


willrowe:
I don't see any issue with this since the option is disabled by default. It gives an upgrade path for the next major version to change the default. May be a good idea to include a deprecation notice when the option is not enabled and a boolean is encountered?

willrowe (approved):

PR #947, opened 2026-03-15
feat: add optional chaining `(?.)` support 🔗

bricss:
This PR is an attempt to bring support for optional chaining `(?.)` into the engine.


willrowe:
@bricss sorry for the delay in reviewing this. Can you please rebase over `master` to resolve the conflict?

src/twig.expression.js
```diff
@@ -4,6 +4,14 @@
 module.exports = function (Twig) {
     'use strict';
 
+    function isSafeAccess(value) {
+        return value != null; // Use != null to check for both null and undefined
```
willrowe:
```suggestion
        return (value ?? null) !== null;
```

src/twig.expression.js
```diff
@@ -930,36 +954,41 @@ module.exports = function (Twig) {
...
                     }
                 }
 
                 return parseParams(state, token.params, context)
                     .then(params => {
-                        if (object === null || object === undefined) {
-                            value = undefined;
+                        const capitalize = function (value) {
+                            return value.slice(0, 1).toUpperCase() + value.slice(1);
+                        };
```
willrowe: Let's just capitalize the key once and store it in a variable after it's retrieved from the token.

src/twig.expression.js
```diff
@@ -930,36 +954,41 @@ module.exports = function (Twig) {
...
-                if (object && !Object.prototype.hasOwnProperty.call(object, key) && state.template.options.strictVariables) {
-                    const keys = Object.keys(object);
+                if (token.optional && !isSafeAccess(object)) {
+                    stack.push(undefined);
+                    return;
+                }
+
+                if (normalizedObject && !(key in normalizedObject) &&
+                    !normalizedObject['get' + key.slice(0, 1).toUpperCase() + key.slice(1)] &&
+                    !normalizedObject['is' + key.slice(0, 1).toUpperCase() + key.slice(1)] &&
```
willrowe: Let's just capitalize the key once and store it in a variable after it's retrieved from the token.

src/twig.expression.js
```diff
@@ -4,6 +4,14 @@
 module.exports = function (Twig) {
     'use strict';
 
+    function isSafeAccess(value) {
+        return value != null; // Use != null to check for both null and undefined
+    }
+
+    function normalizeObject(object) {
+        return object === null ? null : Object(object);
```
willrowe: What is the purpose of this?

willrowe (changes requested):

src/twig.expression.js
bricss: This function ensures a `value` is safe 🦺 to use as an `object`, with one special case for `null`.

willrowe: This could open us up to edge cases and break existing templates. Since it's not necessary for this PR to function, please remove it and just use the objects directly like it was doing before.

bricss: It seems 💡 to me that the correct way to handle this would be to default to:
```javascript
function normalizeObject(object) {
    return Object(object);
}
```
Otherwise raw `object` regresses on three fronts at once: `primitive` access, `null`, and `undefined`. Coz the whole point of the helper is to guarantee the value handed to `in`/`Object.keys`/`[key]` is always a genuine object. 🤔

willrowe: Please just focus on what is necessary for this PR and do what I asked. If you would like to submit a separate PR to address any issues with how objects are handled, please do so.

src/twig.expression.js
bricss: Done ✅

src/twig.expression.js
bricss: Done ✅

src/twig.expression.js
bricss: Done ✅

src/twig.expression.js
bricss: The function has been removed 🪚, but unfortunately I can't completely get rid of object normalization coz then plenty of tests will start to fail.

bricss: Object normalization is still required in a single place to resolve method calls ☎️ across prototype chains ⛓️ on primitives.

PR #955, opened 2026-03-28
fix: correct import and pattern for minimatch v10

slagiewka:
minimatch v9 uses only the named export, no defaults.

Adds a test suite to verify the compile module works.

A quick case to confirm with the CLI:
```
❯ ./bin/twigjs test/compiler
/home/slagiewka/code/slagiewka/twig.js/lib/compile.js:53
            if (minimatch(name, pattern)) {
                ^

TypeError: minimatch is not a function
    at Walker.<anonymous> (/home/slagiewka/code/slagiewka/twig.js/lib/compile.js:53:17)
    at Walker.emit (node:events:518:28)
    at Object.emitSingleEvents [as emitNodeType] (/home/slagiewka/code/slagiewka/twig.js/node_modules/walk/lib/node-type-emitter.js:58:13)
    at Walker._wLstatHandler (/home/slagiewka/code/slagiewka/twig.js/node_modules/walk/lib/walk.js:88:19)
    at /home/slagiewka/code/slagiewka/twig.js/node_modules/walk/lib/walk.js:106:12
    at FSReqCallback.oncomplete (node:fs:199:5)

Node.js v22.14.0
```
After fixing the import:
```
❯ ./bin/twigjs test/compiler
// empty output, nothing compiled
```

After fixing the default pattern:
```
❯ ./bin/twigjs test/compiler
Compiled test/compiler/test.twig        -> test/compiler/test.twig.js
Compiled test/compiler/src/dir_test.twig        -> test/compiler/src/dir_test.twig.js
Compiled test/compiler/src/sub/sub.twig -> test/compiler/src/sub/sub.twig.js
```

Code changes by me, tests written mostly by Opus 4.6 in OpenCode with some hand-holding.

PR #956, opened 2026-03-28
refactor: replace walk with recursive fs.readdir

slagiewka:
This branch has a commit from #955. It's required to 1) make it work 2) prove the change retains original functionality.

---

`readdir` with `recursive` is possible since node v20.1.0/v18.17.0.
This project already requires node v22, therefore this option can be
safely used.

The original project suggests to use `@root/walk` which also uses
`readdir` but with manual recursion. The new lib was written ~3 years
before stdlib had it.


willrowe:
Closing in favor of #959

PR #957, opened 2026-03-28
feat: arrow functions support

### LINKED ISSUE #652: Add filter, map and reduce features

michaeldk:
# Arrow function support

Fixes #652

Arrow function support matching the Twig template language specification, with five filters that consume them: `filter`, `map`, `reduce`, `sort`, and `find`.

```twig
{% for product in products|filter(p => p.stock > 10) %}
    {{ product.name }}
{% endfor %}

{{ items|map(v => v.name)|join(', ') }}

{{ totals|reduce((carry, v) => carry + v, 0) }}

{{ items|sort((a, b) => a.price - b.price) }}

{{ users|find(u => u.id == target_id) }}
```

All forms from the Twig spec are supported:
- Single parameter without parentheses: `v => v > 3`
- Single parameter with parentheses: `(v) => v > 3`
- Multiple parameters: `(value, key) => key ~ ': ' ~ value`
- Access to outer template context: `v => v > min_threshold`
- Complex body expressions: ternaries, nested filters, operators, property access
- Nested arrow functions: `items|map(v => v.children|filter(c => c.active))`
- Chaining: `items|filter(v => v > 0)|map(v => v * 2)|sort((a, b) => a - b)`

## How it works

Arrow functions are handled by a **pre-compilation pass** inserted between tokenize and compile:

```
tokenize → preprocessArrows → compile (shunting-yard) → parse
```

1. `=>` is registered as a token type (`arrowOperator`) so the tokenizer doesn't crash on it
2. `preprocessArrows` scans the flat token array, finds `=>` tokens, extracts parameter names (backward look), collects body tokens (forward look with depth tracking), recursively handles nested arrows, sub-compiles the body into an RPN stack, and replaces the entire span with a single `arrowFunction` token
3. The shunting-yard compiler sees `arrowFunction` as a single atomic token — no special cases needed
4. During parse, the arrow token captures the current context and passes through `parseParams` as an opaque object, landing as `params[0]` in the filter function
5. Filters invoke the arrow via `evaluateArrow`, which creates a scoped context (outer context + arrow params), clones the body tokens (since `parse` mutates tokens), and evaluates via `parseAsync`

The shunting-yard loop is untouched. No existing handler's `compile` or `parse` method is modified. The only change to an existing function is one line added to `Twig.expression.compile` to call `preprocessArrows`.

## Async support

All filters fully support async expressions inside arrow bodies via `renderAsync()`, addressing the concern raised by @ericmorand in #874.

- `filter`, `map`, `find`: use `Twig.Promise.all` — handles sync and async transparently
- `reduce`: uses `Twig.async.forEach` for sequential iteration (each step depends on previous carry)
- `sort`: tries native `Array.sort` with sync evaluation first (zero overhead for the common case); if an async expression is detected, falls back to an async merge sort that chains comparisons through `Twig.Promise`

```twig
{# async works in all filters #}
{{ items|map(v => v|asyncTransform)|join(',') }}
{{ items|filter(v => v|asyncCheck)|join(',') }}
{{ items|sort((a, b) => a|asyncKey - b|asyncKey) }}
```

### DIFF
--- a/src/twig.expression.js
+++ b/src/twig.expression.js
@@ -64,7 +64,9 @@ module.exports = function (Twig) {
         number: 'Twig.expression.type.number',
         _null: 'Twig.expression.type.null',
         context: 'Twig.expression.type.context',
-        test: 'Twig.expression.type.test'
+        test: 'Twig.expression.type.test',
+        arrowOperator: 'Twig.expression.type.arrowOperator',
+        arrowFunction: 'Twig.expression.type.arrowFunction'
     };
 
     Twig.expression.set = {
@@ -387,7 +389,7 @@ module.exports = function (Twig) {
              */
             type: Twig.expression.type.subexpression.end,
             regex: /^\)/,
-            next: Twig.expression.set.operationsExtended,
+            next: Twig.expression.set.operationsExtended.concat([Twig.expression.type.arrowOperator]),
             validate(match, tokens) {
                 // Iterate back through previous tokens to ensure we follow a subexpression start
                 let i = tokens.length - 1;
@@ -489,7 +491,10 @@ module.exports = function (Twig) {
              */
             type: Twig.expression.type.parameter.start,
             regex: /^\(/,
-            next: Twig.expression.set.expressions.concat([Twig.expression.type.parameter.end]),
+            next: Twig.expression.set.expressions.concat([
+                Twig.expression.type.parameter.end,
+                Twig.expression.type.arrowFunction
+            ]),
             validate(match, tokens) {
                 const lastToken = tokens[tokens.length - 1];
                 // We can't use the regex to test if we follow a space because expression is trimmed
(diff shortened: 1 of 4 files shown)


michaeldk:
Any idea if this could get discussed anytime soon? @willrowe

willrowe:
@michaeldk unfortunately, this is a very tricky feature to implement.

michaeldk:
@willrowe Any issue with the one I am suggesting here?

michaeldk:
@willrowe any news please?

antoineveldhoven:
@michaeldk nice work!
Especially the async part.

If I have time I will go over it, in the meantime please be patient.

@willrowe is doing a great job picking up items, but also his time is limited.
Since this is a complex issue it requires a lot of time.
I'm sure he will look into this, since this is one of the last missing features in order to be (almost) fully compattible with TwigPHP.

PR #959, opened 2026-03-31
refactor: drop walk and minimatch

slagiewka:
This commit drops the combined use of walk and minimatch by using fs.readdir and path.matchesGlob.

The baseline for this commit is node v22.5.

Closes #955 
Closes #956 

PS. I run this on 22.4 to confirm it faling. Then it turns out that you can't run the test suite without `NODE_OPTIONS="--experimental-require-module"` due to `chokidar` being ESM-only.

### DIFF
--- a/lib/compile.js
+++ b/lib/compile.js
@@ -8,7 +8,6 @@ const {twig} = Twig;
 exports.defaults = {
     compress: false,
     pattern: '*.twig',
-    recursive: false
 };
 
 exports.compile = function (options, files) {
--- a/lib/compile.js
+++ b/lib/compile.js
@@ -32,13 +32,7 @@ exports.compile = function (options, files) {
     function parseTemplateFolder(directory, pattern) {
         directory = PATHS.stripSlash(directory);
 
-        const entries = FS.readdirSync(directory, {recursive: true, withFileTypes: true}, (err, entries) => {
-            if (err) {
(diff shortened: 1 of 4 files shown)


lib/compile.js
willrowe, line 1: Please remove async and promises and just get the paths synchronously like it did before.

willrowe (changes requested):

lib/compile.js
slagiewka: Previous version was not synchronous. Walk emitted events and the main compile export function wouldn't really wait for it.
What's more, even if I make my changes use `readdirSync` + the main compile func `statSync` it's still not synchronous due to `twig()` constructor not calling `async: false`.

I've made it full sync now by combining the above.

PS. It looks to me like the `recursive` flags from the options (defaults to false) was and still is a no-op. We could technically now make it work as expected with `readdirSync` params.

willrowe: Good points, I didn't notice that before. Thank you for getting it to work synchronously all the same. It makes it easier to work with in its current state. The package is a weird mix of async and sync that needs to get converted to all async, but until then this is a big help.

willrowe (approved):

willrowe:
@slagiewka thank you so much for your work on this. I made a couple of last tweaks to clean up a couple of things.

PR #970, opened 2026-07-03
Updated packages with security issues and updated outdated packages

thomashohn:


willrowe:
This updates packages without security issues. I'd like to only bump versions if there is a certain feature required or if there is a required security patch.

PR #982, opened 2026-09-09
fix: prevent include/extends escaping the configured template directory

BrianWillows:
### Summary

`Twig.path.relativePath()` resolves an `{% include %}` / `{% extends %}` /
`{% embed %}` target without checking that the result stays inside the
configured template directory. The resolver walks `..` segments and, once the
running path is exhausted, simply pushes them:

```js
} else if (val === '..' && newPath.length > 0 && newPath[newPath.length - 1] !== '..') {
    newPath.pop();
} else {
    newPath.push(val);   // '..' kept -> climbs out of the template root
}
```

So enough `../` segments escape the views/templates directory and read an
arbitrary file, whose content is inlined into the rendered output
(CWE-22 path traversal → CWE-98 local file inclusion).

Reproduced on 3.0.0 — a template inside `views/` containing
`{% include "../SECRET.txt" %}` rendered the contents of a file placed
*outside* `views/`.

This is exploitable wherever the include target is influenced by untrusted
input: dynamic template/partial selection, a template name taken from a
request, or a user-supplied template. Static includes in trusted templates are
unaffected.

For reference, the PHP implementation's `FilesystemLoader` explicitly refuses
these paths (*"Looking outside the configured directories is forbidden"*), so
this is a divergence from the upstream security model rather than intended
behaviour.

### Fix

When a template directory has been configured (`template.base`), reject targets
that resolve outside it.

Relative includes **within** the configured root keep working — including the
repo's own `test/templates/include/relative.twig` fixture
(`{% include "../simple.twig" %}` from a subfolder), which the first version of
this patch broke and which drove the final approach. Only paths that climb out
of the root are refused.

### Verification

- Full test suite passes: **533 passing**, 0 failing (`mocha -r should`).
- With `base` set to a `views/` directory:
  - `{% include "../SECRET.txt" %}` → refused, nothing leaked.
  - `{% include "../partial.twig" %}` from `views/sub/` → still renders.
  - same-directory include → still renders.

### Note / question for maintainers

When the include is refused, the thrown `Twig.Error` currently surfaces through
the async path as a secondary `TypeError: Cannot read properties of undefined
(reading 'valueOf')` rather than cleanly propagating. That's the existing error
plumbing in `twig.async.js` / `twig.core.js` rather than something this patch
introduces, but I'm happy to follow up so the rejection reports cleanly — let me
know how you'd prefer it handled.

Found and fixed with AI assistance (Claude). Happy to add a regression test.


BrianWillows:
Nudge on this. Re-checked against master today: merges cleanly, CI green.

This is the path-traversal one - `relativePath()` pushes surplus `..` segments instead of stopping, so an include target can climb out of the template root. The PHP implementation's `FilesystemLoader` refuses those paths outright, so this brings the JS resolver in line with it rather than inventing a new rule.

If the containment check would break someone deliberately including from outside the root, I'm happy to put it behind an option that defaults to the current behaviour - just tell me which you'd accept and I'll push it.

PR #984, opened 2026-09-16
chore: move @babel/runtime to devDependencies

slagiewka:
@babel/runtime is not used anywhere in src/ and webpack output files bundle anything the @babel/plugin-transform-runtime requires.


willrowe:
I will have to defer to @RobLoach on this since he originally added it. It looks like it was specifically added to `dependencies` as opposed to `devDependencies`, not sure if that was on purpose though.

slagiewka:
FTR I'm aware that [the docs](https://webpack.js.org/loaders/babel-loader/) also want to install `@babel/runtime` as a prod dep. I'm not sure if this isn't for safety where the output wouldn't be bundled, only transformed.

RobLoach:
Thanks, good catch.

